PR #10, opened 2021-07-15
Fix nonce & tombstone message + query

arj03:
Bundled two changes into 1 PR. Sorry ;)

The nonce is the recent change to just embed it directly as a buffer since we have a binary format and the second I stumbled upon where the tombstone message was incorrect.

### DIFF
--- a/messages.js
+++ b/messages.js
@@ -1,6 +1,6 @@
 const crypto = require('crypto')
 const bb = require('ssb-bendy-butt')
-const { author, and, type } = require('ssb-db2/operators')
+const { author, and, type, slowEqual } = require('ssb-db2/operators')
 const keys = require('./keys')
 
 // FIXME: define and use json schema
@@ -36,13 +36,13 @@ exports.init = function (sbot) {
     return msgVal
   }
 
-  function getBase64Nonce() {
-    return crypto.randomBytes(32).toString('base64')
+  function getNonce() {
+    return crypto.randomBytes(32)
   }
 
   return {
     addExistingFeed(metafeedKeys, previous, feedpurpose, feedKeys, metadata) {
-      const nonce = getBase64Nonce()
+      const nonce = getNonce()
       return add(feedpurpose, nonce, previous, feedKeys, metafeedKeys, metadata)
     },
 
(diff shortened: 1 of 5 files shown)


arj03:
Will do a few more modification before this is ready

messages.js
```diff
@@ -49,20 +49,24 @@ exports.init = function(sbot) {
...
-      let query = and(author(metafeedKeys.id), type('metafeed/add'))
+      let query = and(
+        author(metafeedKeys.id),
+        type('metafeed/add')
+      )
 
       // FIXME: getJITDB() is not a public API
       sbot.db.getJITDB().all(query, 0, false, false, (err, results) => {
         if (err) return cb(err)
+        results = results.filter(x => x.value.content.subfeed === feedKeys.id)
```
arj03, line 60: In a way it would be great if we had a simple operator in jitdb like slowEqual that we could use for this kind of filter where we have 90% of the query covered by indexes already.

messages.js
```diff
@@ -26,19 +26,19 @@ exports.init = function(sbot) {
                      previous ? previous.value.sequence+1 : 1, +new Date())
   }
 
-  function getBase64Nonce() {
-    return crypto.randomBytes(32).toString('base64')
+  function getNonce() {
+    return crypto.randomBytes(32)
```
staltz, line 30: Isn't it a problem to *encode* Buffers into BFE if we can't *decode* them out? https://github.com/ssb-ngi-pointer/ssb-bfe/blob/158c75b827804f2316019e476f03c033130c18f2/index.js#L223

arj03: Ahh, this is running against an ooold version of db2. Hmm

staltz:
We spoke on a call that this PR should be rebased onto another branch (in my local env) that updates ssb-bendy-butt and db2. So let's draft this one for now. I can do the rebasing later.

staltz:
Tests in CI fail randomly. Once I ran and it failed with node 16.x, the second ran it failed with node 12.x but passed with 16.x.

I'll take that as a sign of a race condition somewhere, unrelated to this PR. Merging

PR #12, opened 2021-07-16
Don't use ssb-db2 internals

staltz:
Removed uses of `ssb.db.onDrain()` and `ssb.db.getJITDB()`, as those are internal APIs.

The diff looks crazy, but just because indentation changed, otherwise it's rather simple PR.


arj03 (approved):
Great, don't know why I used jitdb directly :D

PR #16, opened 2021-07-20
Small refactors

staltz:
Easiest to review commit-by-commit.


arj03 (approved):
Yep, looks good. Read through all of them, simple and to the point. Would be nice to have that restart error fixed ;)

staltz:
Yeah, I took a quick look at the restart issue but a quick look was not enough to understand anything. So it needs a close look

PR #17, opened 2021-07-30
New public API with find / filter / create / findOrCreate

staltz:
- [x] find
- [x] filter
- [x] create
- [x] findOrCreate
- [x] move low-level API docs to inline JS
- [x] update README
- [x] simplify pull-stream chain in getOrCreateMetafeed
- [x] Should we do `ssb.metafeeds.create(cb)` as a shortcut for `ssb.metafeeds.create(null, null, cb)`?
- [x] Should probably rename `{feedpurpose, subfeed, keys}` to `{purpose, id, keys}`?


staltz:
@arj03 Ready for a review although I'll work on the refactor "simplify pull-stream chain in getOrCreateMetafeed". Also I would love to hear your thoughts on:

> * Should we do `ssb.metafeeds.create(cb)` as a shortcut for `ssb.metafeeds.create(null, null, cb)`?
> * Should probably rename `{feedpurpose, subfeed, keys}` to `{purpose, id, keys}`?

README.md
```diff
@@ -3,65 +3,180 @@
...
+
+The response is delivered to the callback `cb`, where the 1st argument is the possible error, and the 2nd argument is the found feed (which can be either the root meta feed `{ seed, keys }` or a sub feed `{ feedpurpose, subfeed, keys }`).
+
+### `sbot.metafeeds.create(metafeed, details, cb)`
+
+*Creates a new subfeed of `metafeed` matching the properties described in `details`.*
+
+`metafeed` can be either `null` or a meta feed object `{ seed, keys }` (as returned by `create()`). If it's null, then the root meta feed will be created, if and only if it does not already exist. If it's null and the root meta feed exists, the root meta feed will be returned via the `cb`.
+
+`details` can be `null` only if if `metafeed` is null, but usually it's an object with the properties:
```
arj03, line 153: if if

README.md
```diff
@@ -3,201 +3,180 @@
...
 
-#### tombstoneFeed(metafeedKeys, previous, feedKeys, reason, cb)
+Notice that each `f` (and the result) is an object that describes the subfeed, with the fields:
 
-Generate a message to be posted on meta feed tombstoning a feed on a
-meta feed. `Previous` is the previous message on the meta feed in key
-value form.
+* `feedpurpose`: same string as used to create the subfeed
+* `subfeed`: the SSB identifier for this subfeed
+* `keys`: the [ssb-keys] compatible `{ curve, public, private, id }` object
```
arj03, line 101: I'll read the code to see what it does, but from just reading the README I was thinking it would also have whatever metadata you have attached as in the example with the game?

README.md
arj03: or only if if => iff :)

README.md
```diff
@@ -3,201 +3,180 @@
...
-sbot.metafeeds.messages.generateAnnounceMsg(metafeedKeys, (err, announceMsg) => {
-  sbot.db.publish(announceMsg, (err) => {
-    console.log("main feed is now linked to meta feed")
-  })
-})
-```
+*Looks for all subfeeds of `metafeed` that satisfy the condition in `visit`.*
 
-#### generateSeedSaveMsg(metafeedId, mainfeedId, seed)
+`metafeed` can be either `null` or a meta feed object `{ seed, keys }` (as returned by `create()`). If it's null, then the result will be an array containing one item, the root meta feed, or zero items if the root meta feed does not exist.
```
arj03, line 131: I can see how this null would be useful in the getOrCreate case. I was thinking at first that null would allow one to search in all metafeeds, not only in your own.

query.js
```diff
@@ -30,13 +32,24 @@ function subfeed(feedId) {
...
+     *   console.log("seed buffer", seed)
+     * })
+     * ```
+     */
     getSeed(cb) {
       // FIXME: maybe use metafeed id
       sbot.db.query(
         where(and(author(sbot.id), type('metafeed/seed'))),
-        toCallback((err, msgs) => {
+        paginate(1),
```
arj03, line 49: I wonder if we do this paginate(1) that we do a descending. If one for whatever reason posts two of those message the latest should be the correct one.

arj03:
That was reading the diff. This looks like a good and simple api.

> Should we do ssb.metafeeds.create(cb) as a shortcut for ssb.metafeeds.create(null, null, cb)?

I think that use case it rather rare in practise as I found think people would call `findOrCreate` anyway for the root meta feed. So no, I don't think we need that. 

> Should probably rename {feedpurpose, subfeed, keys} to {purpose, id, keys}?

I kind of prefer them to be similar to the spec and the fields on the actual messages.

staltz:
I'll comment on the other things later (I'm on mobile), but what do you think about the shortcut `findOrCreate(cb)` equals `findOrCreate(null, null, null, cb)`?

arj03:
I like that shortcut :+1:

README.md
staltz: Done!

README.md
staltz: Oh, that's a feature that didn't even cross my mind when working on this module. Do you see use cases for that?

I think we're going to interact with other people's meta feeds in different modules. Like the replication-scheduler, or whatever "ssb-friends for meta feeds" alternative we come up, that's where I would expect to want to get other people's meta feeds. I imagined that ssb-meta-feeds is kind of like ssb-identities, handling only your own stuff, where `findOrCreate` is the core use case, and these `find`/`filter` alternatives are there just in case the developer knows what they're doing (e.g. if they want to check if the subfeed exists without creating it in case it doesn't).

arj03: I was thinking of the case where if we do indexes of other feeds, then you might have a rule that there should be at least 3 of each indexes as one of the index writers might be offline for an extended period of time. In this case it would be good to be able to query for non-tombstoned indexes in all meta feeds you know of. But I think we should tackle that use case when / if we run into it.

staltz:
Ok, updated this PR!

* fix `if if` typo
* use `descending()` after `paginate(1)`
* hydrate adds `metadata` field too
* support calling APIs with just cb arg

> > Should probably rename {feedpurpose, subfeed, keys} to {purpose, id, keys}?
> 
> I kind of prefer them to be similar to the spec and the fields on the actual messages.

Alright, kept them the same.

Another thing which we may need to address soon is retrieving tombstoned feeds. Currently, they don't appear anywhere in the public API. I was thinking that we could have `findTombstoned` and `filterTombstoned`, but open to feedback on that too.

arj03:
Changes look good. I don't have a better idea than the mirrored api for tombstoned feeds.

README.md
staltz: Yeah, this API seems flexible to allow that in the future

staltz:
Okay, ready for a final review :)

arj03:
I checked the last 3 commits, looks great!

Issue #6, opened 2021-07-07
High level validation


arj03:
This library needs to implement the content validation part of https://github.com/ssb-ngi-pointer/bendy-butt-spec#validation

mycognosist:
Content validation has been merged : https://github.com/ssb-ngi-pointer/ssb-meta-feeds/pull/18

Now it can be utilised in other parts of the module.

PR #18, opened 2021-08-05
Add validation for single meta feeds message

mycognosist:
- [x] validate a single meta feed msg (according to [validation criteria in spec](https://github.com/ssb-ngi-pointer/ssb-meta-feed-spec#usage-of-bendy-butt-feed-format))
- [x] write basic tests
- [x] write tests for `testvector-metafeed-bad-content.json` entries (4/7)
- [x] write documentation for `validateSingle`

-----

`validateSingle(contentSection)` performs the following checks (in order):

- `contentSection` is an array of 2 items
- `content` is an array of 3 or 4 items (`nonce` is optional)
- `content.type` is one of the four valid `metafeed/____` variants (`add`, `update`, `seed` or `tombstone`)
- `content.subfeed` begins with `00` when bfe-encoded
- `content.metafeed` begins with `0003` when bfe-encoded
- bb-encoded `content.nonce` is 35 bytes (check only performed if `content.type === metafeed/add`)
- `contentSignature` uses the `subfeed`'s keypair


arj03:
Fixes #6

validate.js
```diff
@@ -0,0 +1,57 @@
+const bb = require('ssb-bendy-butt')
+const bfe = require('ssb-bfe')
+
+exports.validateSingle = function (contentSection) {
+  if (!Array.isArray(contentSection) || contentSection.length !== 2)
```
arj03, line 5: Another way to write this is: `!(Array.isArray(contentSection) && contentSection.length === 2)`.

validate.js
```diff
@@ -0,0 +1,57 @@
...
+    return new Error(
+      `invalid message: contentSection ${typeof contentSection} with length ${
+        contentSection.length
```
arj03, line 8: I guess this will blow up if contentSection is null or undefined?

validate.js
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+  // bencoded nonce should be 35 bytes: 2 bytes for length, 1 for ':' and 32 for data
+  if (content.type === 'metafeed/add') {
+    const nonceBB = bb.encode(content.nonce)
+    if (nonceBB.length !== 35)
+      return new Error(
+        `invalid message: content nonce is ${nonceBB.length} bytes when bencoded, expected 35`
+      )
+  }
```
arj03, line 56: Should probably return true or similar if the item was validated

validate.js
```diff
@@ -0,0 +1,57 @@
...
+  const [content, contentSignature] = contentSection
+
+  if (!Array.isArray(content) || content.length < 3 || content.length > 4)
+    return new Error(
+      `invalid message: content ${typeof contentSection} with length ${
+        content.length
+      } is incorrect, expected a list of type, subfeed, metafeed and an optional nonce`
+    )
+
+  // TODO: might need to check if content is encrypted?
```
arj03, line 21: I was thinking about this as well on the first check, at first I thought you would not get them, but then thinking where we need this. See [hydrate](https://github.com/ssb-ngi-pointer/ssb-meta-feeds/blob/5cbf43c4201d6a251832afde31b1d5c0f35abce7/query.js#L72) where encrypted content signals that it is not for us so should be discarded. For `publish` like we talked about I would assume we use this function before encrypting.

arj03: Maybe be good this as the first check? To make the error better than the first check.

validate.js
```diff
@@ -0,0 +1,57 @@
...
+
+  if (
+    !(
+      content.type === 'metafeed/add' ||
+      content.type === 'metafeed/update' ||
+      content.type === 'metafeed/seed' ||
```
arj03, line 27: see https://github.com/ssb-ngi-pointer/ssb-meta-feed-spec/pull/20

validate.js
```diff
@@ -0,0 +1,57 @@
...
+  if (!Array.isArray(contentSection) || contentSection.length !== 2)
+    return new Error(
+      `invalid message: contentSection ${typeof contentSection} with length ${
+        contentSection.length
+      } is incorrect, expected a list of content and contentSignature`
+    )
+
+  const [content, contentSignature] = contentSection
+
+  if (!Array.isArray(content) || content.length < 3 || content.length > 4)
```
arj03, line 14: you can have really any kind of metadata on the feed. So instead of checking these lengths I think we should just check that if this is an add then it at least has the required fields, same with tombstone.

validate.js
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+  const metafeedBFE = bfe.encode(content.metafeed)
+  const metafeedType = metafeedBFE.slice(0, 2).toString('hex')
+  if (metafeedType !== '0003')
+    return new Error(
+      `invalid message: content metafeed type "0x${metafeedType}" is incorrect, expected 0x0003`
+    )
+
+  // bencoded nonce should be 35 bytes: 2 bytes for length, 1 for ':' and 32 for data
```
arj03, line 49: this should be just 32 bytes of random data. Buffer is native in bencode so doesn't have to be encoded

arj03:
Overall this looks good, I think this has the right structure. It's good that this spotted an error in the spec so we can get that fixed :)

validate.js
mycognosist: Could you describe how the check for encrypted content would be performed? Would you check `typeof` `contentSection` for `string`?

In the `hydrate` example it appears that encrypted content is implicitly ignored because it is not matched by any of the `.filter()` statements. Do I understand that correctly?

arj03: That is correct, I was thinking we would add the validate check before those filters (line 77) so that we are sure that only valid things are passed further along as that makes checks easier. And yes a typeof check should do the trick.

validate.js
```diff
@@ -0,0 +1,63 @@
...
+ */
+exports.validateSingle = function (contentSection) {
+  if (contentSection === null || contentSection === undefined)
+    return new Error(
+      `invalid message: contentSection cannot be null or undefined`
+    )
+
+  // check if content is (maybe) encrypted
+  if (typeof contentSection === 'string')
+    return new Error(
```
mycognosist, line 17: I'm not sure an `Error` return type makes the most sense here but I also don't want to make the return handling unnecessarily complex.

My thinking is that the caller checks the return type: if `true`, all good; if `Error`, check the `err.message` and act accordingly (could be used to catch encrypted input case).

mycognosist:
I've added signature verification but am not 100% sure I've got it right. @arj03 , could you please take a look?

arj03:
> I've added signature verification but am not 100% sure I've got it right. @arj03 , could you please take a look?

Looks good from what I can see.

mycognosist:
Please ignore the failed CI tests for now. I'm working on it.

validate.js
```diff
@@ -0,0 +1,155 @@
...
+  const isSignatureRx = isCanonicalBase64('', '\\.sig.\\w+')
+
+  if (!isSignatureRx.test(contentSignature))
+    return new Error(
+      `invalid message: contentSignature "${contentSignature}", expected a base64 string`
+    )
+
+  const contentBFE = bfe.encode(content)
+
+  // if the subfeedKey is a supported uri, convert it to sigil for verification
```
mycognosist, line 102: @staltz 

This is my attempted solution of the 'subfeed key might be a URI' case.

staltz: Looking good

validate.js
```diff
@@ -0,0 +1,155 @@
...
+    if (
+      !SSBURI.isFeedSSBURI(subfeedKey) &&
+      !SSBURI.isBendyButtV1FeedSSBURI(subfeedKey) &&
+      !SSBURI.isGabbyGroveV1FeedSSBURI(subfeedKey)
+    ) {
+      return new Error(
+        `invalid message: subfeed key "${subfeedKey}", expected a canonical uri format or classic ssb sigil`
+      )
+    } else {
+      let { type, format, data } = SSBURI.decompose(subfeedKey)
```
staltz, line 113:
```suggestion
      const { data } = SSBURI.decompose(subfeedKey)
```

staltz: We actually don't need the sigil and the suffix, because on line 120 we're passing it to `public`, and `public` consists of only the base64 data.

mycognosist: That's much neater, thanks! Wasn't quite sure how to do the selective destructuring 👍🏻 👍🏻

validate.js
```diff
@@ -0,0 +1,156 @@
...
+  if (hmacKeyErr) return hmacKeyErr
+
+  const isSignatureRx = isCanonicalBase64('', '\\.sig.\\w+')
+
+  if (!isSignatureRx.test(contentSignature))
+    return new Error(
+      `invalid message: contentSignature "${contentSignature}", expected a base64 string`
+    )
+
+  const contentBFE = bfe.encode(content)
```
arj03, line 102: maybe move this down to after the subfeed check to make it easier to read

arj03 (approved):
Just found a minor stylistic thing. Otherwise looks great I think. 

I think I'm okay with tests only being based on test vectors. We need to add validation in this module in places like query, so that will act as testing in a way. But that can be done in another PR. Might be good to have this PR as just adding validation methods on its own.

staltz (approved):
Looks good.

Any idea, @arj03 or @mycognosist, where this new API will be called? In `ssb-meta-feeds` or in `ssb-db2`?

arj03:
I would think that it needs to be used in this repo as mentioned in my review. Both for internal testing but also if go or another implementation created the metafeed and we run some JS on top of that. Other than that I think we need to use this validation in https://github.com/ssb-ngi-pointer/ssb-replication-scheduler/pull/5 because it touches other peoples meta feed. We don't need this in db2 I think because db2 should not care about the "content", the base validation is done there instead because we cannot reject a message coming over the network with invalid content because then we would break the feed.

arj03:
It would be good if we do the validate before we call the `sbot.db.add` as well in this repo

PR #21, opened 2021-08-09
hacky fix for race condition in tests

### LINKED ISSUE #19: Irregular test errors
Test failures occasionally occur during CI runs. I have not yet encountered these errors while testing on my machine.

Here's an [example failure](https://github.com/ssb-ngi-pointer/ssb-meta-feeds/pull/18/checks?check_run_id=3277963806) for the validation PR.

```shell
TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined

    at Function.from (buffer.js:330:9)
    at /home/runner/work/ssb-meta-feeds/ssb-meta-feeds/query.js:57:34
```

It seems that the `getSeed()` query ([L57 in `query.js`](https://github.com/ssb-ngi-pointer/ssb-meta-feeds/blob/master/query.js#L57)) is being passed an `undefined` value:

`const seedBuf = Buffer.from(msg.value.content.seed, 'hex')`

The `getSeed()` function is called from [L119 of `test/query.js`](https://github.com/ssb-ngi-pointer/ssb-meta-feeds/blob/master/test/query.js#L119).

staltz:
Related to #19 but shouldn't close #19

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -153,23 +153,30 @@ test('restart sbot', (t) => {
         path: dir,
       })
 
-    sbot.metafeeds.findOrCreate(null, null, {}, (err, mf) => {
-      t.error(err, 'no err')
-      t.ok(Buffer.isBuffer(mf.seed), 'has seed')
-      t.ok(mf.keys.id.endsWith('.bbfeed-v1'), 'has key')
-
-      sbot.metafeeds.filter(mf, null, (err, filtered) => {
+    // The setTimeout is a hacky solution for a race condition in ssb-db2.
+    // findOrCreate is going to look up for the "seed" message, which is a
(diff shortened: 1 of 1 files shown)


staltz:
I'm getting a different error in CI now: 

```
/home/runner/work/ssb-meta-feeds/ssb-meta-feeds/node_modules/ssb-bfe/index.js:148
    else throw new Error('Unknown feed: ' + buf)
         ^

Error: Unknown feed:
```

mycognosist (approved):
🟢

mycognosist:
> I'm getting a different error in CI now:

Oh yeah I saw this one last week as well.

PR #22, opened 2021-08-09
Adding existing feed should have no nonce

staltz:
Related to https://github.com/ssb-ngi-pointer/ssb-meta-feed-spec/issues/22

1st test :x: , 2nd fix :heavy_check_mark:

### DIFF
--- a/query.js
+++ b/query.js
@@ -141,7 +141,6 @@ exports.init = function (sbot, config) {
     hydrateFromMsg(msg, seed) {
       const { feedpurpose, subfeed, nonce } = msg.value.content
       const metadata = self.collectMetadata(msg)
-      const nonceB64 = nonce.toString('base64')
       const feedformat = subfeed.endsWith('.bbfeed-v1')
         ? 'bendy butt'
         : 'classic'
@@ -150,7 +149,7 @@ exports.init = function (sbot, config) {
           ? config.keys
           : sbot.metafeeds.keys.deriveFeedKeyFromSeed(
               seed,
-              nonceB64,
+              nonce.toString('base64'),
               feedformat
             )
       return {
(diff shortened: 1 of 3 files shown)


query.js
```diff
@@ -141,11 +141,13 @@ exports.init = function (sbot, config) {
     hydrateFromMsg(msg, seed) {
       const { feedpurpose, subfeed, nonce } = msg.value.content
       const metadata = self.collectMetadata(msg)
-      const nonceB64 = nonce.toString('base64')
       const keys =
         subfeed === sbot.id
           ? config.keys
-          : sbot.metafeeds.keys.deriveFeedKeyFromSeed(seed, nonceB64)
+          : sbot.metafeeds.keys.deriveFeedKeyFromSeed(
```
mycognosist, line 147: So this (`hydrateFromMsg()`) will only ever get called with a `msg` that was generated for a new SSB identity? In other words, `nonce` will never be `null` or `undefined` for the L149 call?

mycognosist (approved):

query.js
staltz: No, this function will be called for both cases, but in the case of an existing feed, the `nonce` is undefined, so we cannot call `nonce.toString('base64')`

mycognosist: That's what I was wondering about. Calling `.toString('base64')` on `null` or `undefined` produces a `TypeError`, right?

staltz: Oh yeah, it might not be visually obvious, but when `subfeed === sbot.id`, then we assume the case "Existing feed".

mycognosist: Ah I understand now. I knew there must be something I was missing. Thanks for explaining.

staltz:
There's a git conflict, I'll resolve that first

PR #26, opened 2021-08-13
change metafeed/add to its two subtypes

staltz:
Here it is, the missing PR. Turns out I had pushed this branch before I formatted my computer's disk.


mycognosist (approved):
Looks good 🟢

PR #27, opened 2021-08-16
replace publishAs() with add() for bendy butt msgs

staltz:
Follow up to https://github.com/ssb-ngi-pointer/ssb-db2/pull/248, I also changed the API names in `sbot.metafeeds.messages` to make it easier to remember:

- `msgValXXXX()` creates a `msgVal` for content.type `XXXX`
- `contentYYYY()` creates a `content` for type `YYYY`

Meta feed `msgVal`s use `db.add()` not `db.publishAs()`.

### DIFF
--- a/api.js
+++ b/api.js
@@ -113,7 +113,7 @@ exports.init = function (sbot, config) {
       if (!details.feedformat) return cb(new Error('Missing feedformat'))
       sbot.metafeeds.query.getLatest(metafeed.keys.id, (err, latest) => {
         if (err) return cb(err)
-        const msgVal = sbot.metafeeds.messages.msgValAddDerived(
+        const msgVal = sbot.metafeeds.messages.getMsgValAddDerived(
           metafeed.keys,
           latest,
           details.feedpurpose,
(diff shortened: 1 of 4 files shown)


api.js
```diff
@@ -154,7 +154,7 @@ exports.init = function (sbot, config) {
     // Pluck relevant internal APIs
     const { deriveRootMetaFeedKeyFromSeed } = sbot.metafeeds.keys
     const { getSeed, getAnnounces, getLatest } = sbot.metafeeds.query
-    const { generateSeedSaveMsg, generateAnnounceMsg, addExistingFeed } =
+    const { contentSeed, contentAnnounce, msgValAddExisting } =
```
arj03, line 157: Only minor thing I have is that in this file it gets a bit harder to read maybe what contentSeed does because we extract the functions from the object. Like the tests are `messages.contentSeed` which makes it a bit easier to see that is has something to with generating messages. One way to maybe improve this would be to have just messages shorted here, so the lines in this file becomes the same as in the tests. What do you think?

arj03 (approved):
Just one minor thing, I'm good either way. Good to see that all of these are internal methods so no README update this time :)

api.js
staltz: Ah yes, I see what you mean. I can do what you said, but how about (new idea) we just prefix all these functions with `get`?

arj03: also works, probably a better idea

staltz:
Thanks for the review!

PR #28, opened 2021-08-19
Bump ssb-db2 & bendy butt

arj03:
Simple bump some dependencies. Most important is the db2 so we can remove that branch.


staltz (approved):
Merge away! :)

PR #30, opened 2021-08-19
Fix rootmeta concurrency

### LINKED ISSUE #29: getOrCreateRootMetafeed is not concurrency safe
This screenshot shows the problem pretty well, 3 meta feed messages and 2 seed values.

image

We need to do something similar to what we did in jitdb with only allowing 1 at a time.

arj03:
Should fix #29. Gonna test it now.

### DIFF
--- a/api.js
+++ b/api.js
@@ -150,7 +150,26 @@ exports.init = function (sbot, config) {
     }
   }
 
+  // concurrent index helpers
+  function onlyOneAtRootMetaATime(cb) {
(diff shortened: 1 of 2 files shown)


arj03:
Note this is on top of the other PR to allow me to test this

arj03:
Yep, seems to work :)

arj03:
This can be simplied, I'll do this properly tomorrow. Enough for today ;)

staltz:
Very well spotted, good fix, conceptually. I made #31 because it was fun to simplify it.

PR #31, opened 2021-08-19
refactor rootMetaFeedLock

staltz:
I was gonna review #30 by suggesting dropping `waitingRootMetaKey`, and then I found a couple other things, so it was easier to make a PR for the PR. :)


arj03:
Great! One variable less than the solution I was thinking about :guitar:

PR #32, opened 2021-08-20
Use SSB URIs

staltz:
- [x] Pending on ssb-bendy-butt https://github.com/ssb-ngi-pointer/ssb-bendy-butt/pull/29
- [x] Tests here should pass
- [x] Pending on ssb-db2 https://github.com/ssb-ngi-pointer/ssb-db2/pull/255

### DIFF
--- a/keys.js
+++ b/keys.js
@@ -1,6 +1,7 @@
 const crypto = require('crypto')
 const ssbKeys = require('ssb-keys')
 const hkdf = require('futoin-hkdf')
+const SSBURI = require('ssb-uri2')
 
 /**
  * Operations related to keys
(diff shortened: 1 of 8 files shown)


staltz:
FYI @mycognosist @arj03 this branch exists

keys.js
```diff
@@ -23,23 +24,23 @@ const keys = {
    * @param {Buffer} seed
    */
   deriveRootMetaFeedKeyFromSeed(seed) {
-    return keys.deriveFeedKeyFromSeed(seed, 'metafeed', 'bendy butt')
+    return keys.deriveFeedKeyFromSeed(seed, 'metafeed', 'bendybutt-v1')
```
staltz, line 27: It makes underlying code easier to handle if the format is `bendybutt-v1` because then it matches exactly what SSB URI and BFE expect, and also I suspect that the space in `bendy butt` could have surprised some developer in the future

query.js
```diff
@@ -159,7 +160,7 @@ exports.init = function (sbot, config) {
         subfeed,
         keys,
         metadata,
-        seed: feedformat === 'bendy butt' ? seed : undefined,
+        seed: feedformat === 'bendybutt-v1' ? seed : undefined,
```
arj03, line 163: This check should be `type === 'metafeed/add/existing'`

arj03: It seems our tests does not catch this. Maybe add a test where you add an existing feed & derived feed and later restore from a seed.

arj03: I can have a go at fixing this

arj03: https://github.com/ssb-ngi-pointer/ssb-meta-feeds/pull/33

arj03 (approved):
Looks good :)

PR #37, opened 2021-09-01
Integrate validation checks

mycognosist:
Now that we have a means of validating the `contentSection` of a meta feeds message, we can begin to use it in this module (and others).

I'm beginning by adding a simple check before `sbot.db.add` is called in the `create` function. I believe this results in the validation check being called on three occasions in the current test suite:

`findOrCreate() a sub feed`
`findOrCreate() a sub meta feed`
`findOrCreate() a subfeed under a sub meta feed`

### DIFF
--- a/validate.js
+++ b/validate.js
@@ -20,6 +20,8 @@ function isValid(msg, hmacKey) {
     const validationResult = validateSingle(contentSection, hmacKey)
 
     return validationResult === undefined
+  } else {
+    return false
   }
 }
 
--- a/validate.js
+++ b/validate.js
@@ -10,11 +10,27 @@ const CONTENT_SIG_PREFIX = Buffer.from('bendybutt', 'utf8')
 /**
  * Validate a single meta feed message.
  *
+ * @param {Object} msg - a meta feed message in the form of a JSON object
+ * @param {Buffer | string | null} hmacKey - a valid HMAC key for signature verification
+ * @returns {true | false} `true` in the case of successful validation
+ */
+function isValid(msg, hmacKey) {
+  if (msg.value.content && msg.value.contentSignature) {
+    const contentSection = [msg.value.content, msg.value.contentSignature]
+    const validationResult = validateSingle(contentSection, hmacKey)
+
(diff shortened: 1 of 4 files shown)


arj03:
Nice to see validation working here :-)

mycognosist:
It's always encouraging when something succeeds on the first attempt :)

mycognosist:
@arj03 

I'm interested to hear your thoughts on the latest commit (validation during hydration); it's basically just a `map()` which only returns valid messages.

I would take the same kind of approach in the `getMetadata()` function in `query.js`.

Other functions in `query` don't seem to require validation, ie. `getSeed()` and `getAnnounces()`; these would fail validation, since we only validate `metafeed/add/existing`, `metafeed/add/derived`, `metafeed/tombstone` and `metafeed/update` content types.

arj03:
Instead of a map I would do a filter. We don't want the unvalid messages. You also don't need the first filter, that should be a concern of the validate function.

You are absolutely correct that we only need to check messages on a meta feed, not on main feed. So not "everything" coming from the database :)

validate.js
```diff
@@ -10,11 +10,27 @@ const CONTENT_SIG_PREFIX = Buffer.from('bendybutt', 'utf8')
...
+ * @param {Buffer | string | null} hmacKey - a valid HMAC key for signature verification
+ * @returns {true | false} `true` in the case of successful validation
+ */
+function isValid(msg, hmacKey) {
+  if (msg.value.content && msg.value.contentSignature) {
+    const contentSection = [msg.value.content, msg.value.contentSignature]
+    const validationResult = validateSingle(contentSection, hmacKey)
+
+    return validationResult === undefined
+  }
```
staltz, line 23: Should we return a boolean in the `else` case?

mycognosist:
I think this might be complete. @arj03 and @staltz , do you notice anything missing? Otherwise I'll go ahead and merge this.

arj03 (approved):
Looks good to me :)

staltz (approved):

PR #39, opened 2021-09-06
Update subfeed prefix offset value (account for SSB URIs)

mycognosist:
Takes SSB URIs for `bendybutt-v1` and `gabbygrove-v1` into account when calculating and assigning the `prefixOffset` for the subfeed `equal()` db2 operator call

In response to: https://github.com/ssb-ngi-pointer/ssb-meta-feeds/issues/38

-----

A question which came up for me while making this PR:

Would it be more efficient (or accurate) to calculate the required `prefixOffset` based on the given `feedId`, rather than using the max length as offset for all cases?

### DIFF
--- a/query.js
+++ b/query.js
@@ -11,6 +11,12 @@ const {
 } = require('ssb-db2/operators')
 const SSBURI = require('ssb-uri2')
 
+const SUBFEED_PREFIX_OFFSET = Math.max(
+  '@'.length,
+  'ssb:feed/bendybutt-v1/'.length,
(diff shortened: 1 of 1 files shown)


staltz:
> Would it be more efficient (or accurate) to calculate the required `prefixOffset` based on the given `feedId`, rather than using the max length as offset for all cases?

It would be actually less efficient (with regards to runtime performance) but I agree that it would be a better/smarter solution and more accurate too. We settled on this compromise for now because the changes in JITDB would be several.

staltz (approved):
Thanks!

PR #40, opened 2021-09-08
Support gabby grove in deriveFeedKeyFromSeed

mycognosist:
Adds a check for `gabbygrove-v1` when deriving a feed key from seed. This ensures that Gabby Grove is treated in the same way as Bendy Butt when deriving the key (preventing the default behaviour, which is to handle it as a classic format).

Related to: https://github.com/ssb-ngi-pointer/ssb-bendy-butt/pull/35#discussion_r703382283

### DIFF
--- a/keys.js
+++ b/keys.js
@@ -38,7 +38,7 @@ const keys = {
    * ```
    * @param {Buffer} seed
    * @param {string} label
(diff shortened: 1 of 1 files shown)


staltz (approved):
:star:

PR #41, opened 2021-09-10
introduce findById()

staltz:
## Context

In the new multi-format ssb-ebt we need to be able to get any feedID and infer what is it: is it a classic `main` feed? Is it a `classic` index feed? Is it a bendy butt? (The first 2 questions are the most important) It's not enough to just look at the feedId and infer the `feedformat`, we also need to know the `feedpurpose`. In ssb-replication-scheduler we will also need something like this.

## Solution

New API `findById()`. Turns out it was already implemented as `getMetadata`, I just did further tweaking in its internals, to:

- If the feedId given to `findById()` has been tombstoned, return `null`
- Infer the `feedformat` from the `feedId`
- Remove msg.value.content stuff such as `type` and `nonce` etc
- Bundle unknown fields in `msg.value.content` under `metadata` object, like it is during subfeed creation

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,9 @@ Add this plugin like this:
 
 ## Example usage
 
-Let's start by creating a **root meta feed**, necessary for using this module. It
-lives alongside your existing "classical" feed, which we'll refer to as **main
-feed**.
+Let's start by creating a **root meta feed**, necessary for using this module.
+It lives alongside your existing "classical" feed, which we'll refer to as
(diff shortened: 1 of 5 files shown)


arj03 (approved):
Great we can reuse the existing function and just extend it a bit. I have some comments on the usage in ebt, but I'll take them in another channel. This PR in itself is really good in that it makes this module quite a bit more general.

PR #42, opened 2021-09-13
add findByIdSync API

staltz:
- [x] Pending on https://github.com/ssb-ngi-pointer/ssb-db2/pull/260

Adds `findByIdSync`, similar to `findById`, but synchronous (see e.g. similar to `fs.readFile` and `fs.readFileSync`). The synchronous variant requires calling `loadState(cb)` before usage, otherwise it throws an error.

I also refactored/moved stuff from query.js to feeds-lookup.js.

### DIFF
--- a/feeds-lookup.js
+++ b/feeds-lookup.js
@@ -110,12 +110,10 @@ exports.init = function (sbot, config) {
           stateLoaded = true
           cb() // signal that we're ready to findByIdSync
 
-          // sbot.close.hook(function (fn, args) {
-          //   console.log('will abort liveDrainer')
-          //   if (liveDrainer) liveDrainer.abort()
-          //   console.log('did abort liveDrainer')
-          //   fn.apply(this, args)
-          // })
+          sbot.close.hook(function (fn, args) {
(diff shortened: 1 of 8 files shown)


feeds-lookup.js
```diff
@@ -0,0 +1,181 @@
...
+    for (const key of keys) {
+      metadata[key] = content[key]
+    }
+    details.metadata = metadata
+    return details
+  }
+
+  function updateLookup(msg) {
+    const content = msg.value.content
+    if (content.type.startsWith('metafeed/add/')) {
```
arj03, line 95: This needs a validate.isValid here

arj03:
Found one thing that needs fixing before merging. Otherwise look good. It's great to see that this can use the general authorIsBendyButtV1. Should be plenty fast for EBT. Nice work.

staltz:
Thanks for taking a look. How about commit https://github.com/ssb-ngi-pointer/ssb-meta-feeds/pull/42/commits/5e76cef3f164894afb12aa96f96a0fa578763d11 ?

arj03 (approved):
Yup, that should work :)

PR #44, opened 2021-09-14
test findById/findByIdSync properly

staltz:
:x: :heavy_check_mark:

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -162,13 +162,19 @@ tape('findOrCreate() a sub meta feed', (t) => {
   })
 })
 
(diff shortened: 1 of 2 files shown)


test/api.js
```diff
@@ -162,13 +162,19 @@ tape('findOrCreate() a sub meta feed', (t) => {
   })
 })
 
+let testRootMF
```
arj03, line 165: seems like this variable is not used?

staltz: cool

arj03 (approved):
Great! :)

PR #45, opened 2021-09-14
metafeed/announce is signed by meta feed

staltz:
For issue https://github.com/ssb-ngi-pointer/ssb-meta-feeds-spec/issues/27

### DIFF
--- a/test/query.js
+++ b/test/query.js
@@ -123,7 +123,7 @@ test('seed', (t) => {
 })
 
 test('announce', (t) => {
-  messages.getContentAnnounce(metafeedKeys.id, (err, content) => {
+  messages.getContentAnnounce(metafeedKeys, (err, content) => {
     db.publish(content, (err, publishedAnnounce) => {
       t.error(err, 'no err')
       sbot.metafeeds.query.getAnnounces((err, announcements) => {
--- a/validate.js
+++ b/validate.js
@@ -210,7 +210,7 @@ function validateMetafeedAnnounce(msg) {
   if (!ssbKeys.verifyObj(ed25519Public, msg.value.content)) {
     return new Error(
       `metafeed/announce ${msg.key} is invalid ` +
-        `because content is not a signed by the meta feed: ${msg.value.content}`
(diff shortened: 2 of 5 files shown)


staltz:
Oh, I should probably update validation to check these sigs.

staltz:
> Oh, I should probably update validation to check these sigs.

Oh well actually `metafeed/announce` is published on the main feed, and we only need to validate remote peer's `metafeed/announce` msgs, which should happen in ssb-replication-scheduler.

So this PR is ready to go I guess

staltz:
Or hey I can just make a new function, `validateAnnounce(msg)`. Yes

validate.js
```diff
@@ -173,5 +182,39 @@ function validateHmacKey(hmacKey) {
...
+        `because content.metafeed is not a bendy butt feed: ${metaFeedId}`
+    )
+  }
+
+  const { data } = SSBURI.decompose(metaFeedId)
+  const ed25519Public = `${data}.ed25519`
+  if (!ssbKeys.verifyObj(ed25519Public, msg.value.content)) {
+    return new Error(
+      `metafeed/announce ${msg.key} is invalid ` +
+        `because content is not a signed by the meta feed: ${msg.value.content}`
```
arj03, line 213: => because content is not signed by the meta feed: ${msg.value.content}

test/query.js
```diff
@@ -123,7 +123,7 @@ test('seed', (t) => {
 })
 
 test('announce', (t) => {
-  messages.getContentAnnounce(metafeedKeys.id, (err, content) => {
+  messages.getContentAnnounce(metafeedKeys, (err, content) => {
```
arj03, line 126: There must have been a bug here before?

arj03 (approved):
Just found that minor spelling thing, otherwise looks good

test/query.js
staltz: Yes there was. The checking is done correctly in `test/messages.js` but in this file we want to test the query APIs so we just used `getContentAnnounce` to create the inputs, but we didn't validate the input. Anyway whatever :)

PR #48, opened 2021-09-16
new API ensureLoaded()

staltz:
## Context

ssb-meta-feeds `findByIdSync` is used in the new ssb-ebt https://github.com/ssbc/ssb-ebt/pull/52 which in turn is used in https://github.com/ssb-ngi-pointer/ssb-replication-scheduler/pull/5 . 

## Problem

We bumped into a race condition where there are two pull.drains on meta feed messages: **(A)** one in ssb-meta-feeds `feeds-lookup.js` (to keep track of the "details" object associated which each (foreign or local) subfeed ID) and **(B)** another one in `ssb-replication-scheduler` to trigger the replication of a subfeed. We realized that the processing in **(A)** for a specific subfeed ID  **must** happen before the respective processing in **(B)** because information calculated in **(A)** will be needed by ssb-ebt once **(B)** calls into ssb-ebt APIs.

## Solution

New API `ensureLoaded(feedId, cb)` similar to `loadState(cb)`, but for a specific `feedId`.

While `loadState(cb)` just kicks off the pull.drain for all possible meta feed msgs, `ensureLoaded()` will wait until the specific `feedId` has been processed by the **next** `updateLookup()`.

This should be one of the pieces of the puzzle to solve the Problem described above, we will also need a few changes in ssb-ebt to fully solve it.

### DIFF
--- a/feeds-lookup.js
+++ b/feeds-lookup.js
@@ -46,7 +46,22 @@ exports.init = function (sbot, config) {
   const stateLoadedP = DeferredPromise()
   let loadStateRequested = false
   let liveDrainer = null
-  const lookup = new Map()
+  const lookup = new Map() // feedId => details
+  const ensureQueue = {
+    _map: new Map(), // feedId => Array<Callback>
+    add(feedId, cb) {
+      if (this._map.has(feedId)) this._map.get(feedId).push(cb)
+      else this._map.set(feedId, [cb])
+    },
+    flush(feedId) {
+      const queue = this._map.get(feedId)
+      this._map.delete(feedId)
+      while (queue && queue.length > 0) {
+        const cb = queue.shift()
+        stateLoadedP.promise.then(cb)
+      }
+    },
+  }
 
   function assertFeedId(feedId) {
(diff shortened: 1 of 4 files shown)


feeds-lookup.js
```diff
@@ -136,7 +153,14 @@ exports.init = function (sbot, config) {
         loadStateRequested = true
         loadState()
       }
-      stateLoadedP.promise.then(cb)
+      if (cb) stateLoadedP.promise.then(cb)
```
arj03, line 156: nice catch

staltz: Oh yeah there was that

arj03 (approved):
Looks good. The only thing I was wondering is if stateLoadedP.promise.then in flush would be better int he beginning, but I think it should be ok where it is.

staltz:
That one is for the case that we call `ensureLoaded` before `loadState` is done, we don't want the `cb`s to be called

arj03:
Right, should work either way I think

PR #49, opened 2021-09-17
rename one "ed25519" to "classic"

staltz:
This one slipped through. For consistency, we should use `classic` everywhere in this module.


mycognosist (approved):
Oh, I didn't know about this change. I definitely prefer `classic`. Nice 🟢

PR #50, opened 2021-09-21
new API branchStream()

staltz:
- [x] implement pull source `branchStream()`
- [x] tests
- [x] Update README
- [x] Hide/deprecate other APIs?

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,23 +128,22 @@ basis to make sure that `findByIdSync` will operate at the correct time when the
 Returns a **[pull-stream] source** of all "branches" in the meta feed trees.
 
 A "branch" is an array where the first item is the root meta feed and the
-subsequent items are the children and grandchildren of the root. A branch looks
-like this:
+subsequent items are the children and grandchildren (and etc) of the root. A
+branch looks like this:
 
 ```js
 [
   [rootMetafeedId, null],
   [childMetafeedId, childDetails],
   [grandchildMetafeedId, grandchildDetails],
-  [grandgrandchildSubfeedId, grandgrandChildDetails],
 ]
 ```
 
-Or in general, an `Array<[FeedId, Details | null]>`. **details** if the object
-with `{feedpurpose, feedformat, metafeed, metadata}` like in `findById`.
+Or in general, an `Array<[FeedId, Details | null]>`. The **details** object has
+the shape `{ feedpurpose, feedformat, metafeed, metadata }` like in `findById`.
 
 `branchStream` will emit all possible branches, which means sub-branches are
-included. For instance, in the example above, `branchStream` would also emit:
+included. For instance, in the example above, `branchStream` would emit:
 
 ```js
 [
--- a/README.md
+++ b/README.md
@@ -30,21 +30,21 @@ It lives alongside your existing "classical" feed, which we'll refer to as
 **main feed**.
 
 ```js
-sbot.metafeeds.create((err, metafeed) => {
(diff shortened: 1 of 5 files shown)


staltz:
This PR is ready for review, and it's stable since I'm using it in https://github.com/ssb-ngi-pointer/ssb-replication-scheduler/pull/5 .

I took a look at all usages of ssb-meta-feeds and it seems like `filter` was never used, and in all cases where `create` is used, we should use `findOrCreate` instead, so I ended up deprecating (just by hiding from the README) those APIs that are not used. I think it's better to build APIs for actual use cases, than to work for imaginary use cases.

Now that we have `loadState`, I think `find` could be repurposed so that it works also on foreign meta feeds. Currently it's a bit confusing that some APIs operate only on your own meta feeds (`find`, `findOrCreate`) while other APIs operate on meta feeds from anyone (`findByIdSync`, `branchStream`). If we make `find` work for any meta feed, then it's obvious that `findOrCreate` is only for your own meta feed due to the name 'create'.

But I could do that change to `find` in a separate PR because this one is about `branchStream`, and it's ready.

README.md
```diff
@@ -163,6 +123,64 @@ You can also call `sbot.metafeeds.ensureLoaded(feedId, cb)` on an individual
...
+```js
+[
+  [rootMetafeedId, null],
+  [childMetafeedId, childDetails],
+  [grandchildMetafeedId, grandchildDetails],
+  [grandgrandchildSubfeedId, grandgrandChildDetails],
+]
+```
+
+Or in general, an `Array<[FeedId, Details | null]>`. **details** if the object
```
arj03, line 143: Something wrong with this sentence. **details** if? Maybe something like: **details** if defined will contain ... like ...

README.md
```diff
@@ -163,6 +123,64 @@ You can also call `sbot.metafeeds.ensureLoaded(feedId, cb)` on an individual
...
+
+and
+
+```js
+[
+  [rootMetafeedId, null],
+  [childMetafeedId, childDetails],
+  [grandchildMetafeedId, grandchildDetails],
+]
+```
```
arj03, line 172: either etc here or remove grandgrandchildSubfeedId from the example above, because that one would also be emitted, right?

README.md
staltz: Ah, `if` => `is`.

README.md
```diff
@@ -279,6 +299,8 @@ arbitrary tombstoned subfeed under `metafeed` is returned.
 The response is delivered to the callback `cb`, where the 1st argument is the
 possible error, and the 2nd argument is the found tombstoned feed.
 
+-->
```
arj03, line 302: Is this some leftover thing from commenting out stuff?

staltz: Nope, that's a comment that is just hiding the deprecated APIs from the README, so that developers using this library try to use the other APIs for now. I didn't want to delete code, not yet

arj03: Ah, nevermind now I see the start.

arj03:
> This PR is ready for review, and it's stable since I'm using it in ssb-ngi-pointer/ssb-replication-scheduler#5 .

This is a bit light on tests for this new api, but since it is used in the other module I guess that counts as testing :) 

> But I could do that change to find in a separate PR because this one is about branchStream, and it's ready.

Yes! It would be great if find worked on any feed.

arj03 (approved):
Lets get this merged and continue working on the API in another branch. Overall it's great to have this kind of API in here because it makes it similar to ssb-friends and other modules.

PR #51, opened 2021-09-23
cache the root meta feed in memory

staltz:
Here's something nice I bumped into.

If the root meta feed has already been fetched (from disk) or created, then we can just store it in memory, in the lock. This should be an optimization because it prevents subsequent calls to look it up from the disk.

(Actually, now that I'm typing this, I'm not sure if this is worthwhile optimization. ssb-db2 and jitdb seem to have quite a good amount of caching already. Feel free to say no to this PR)


arj03 (approved):
Not at all, this is a great optimization!

PR #55, opened 2021-10-06
branchStream supports opts.tombstoned=null

staltz:
Currently, `branchStream` supports `opts.tombstoned` as a boolean. 

In ssb-replication-scheduler I need a way of looking up the *non-tombstoned* subfeeds of a *tombstoned* feed, because those subfeeds are to be considered implicitly tombstoned as well. The current `branchStream` doesn't let me do that.

This PR changes `opts.tombstoned` to allow `null`, which means I can look up subfeeds regardless of their tombstoning status.

### DIFF
--- a/feeds-lookup.js
+++ b/feeds-lookup.js
@@ -273,7 +273,7 @@ exports.init = function (sbot, config) {
         live = true,
         old = false,
         root = null,
-        tombstoned = false,
+        tombstoned = null,
       } = opts || {}
 
       const filterRootFn = root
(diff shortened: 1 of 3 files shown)


feeds-lookup.js
```diff
@@ -273,19 +273,21 @@ exports.init = function (sbot, config) {
...
         root = null,
-        tombstoned = false,
+        tombstoned = null,
       } = opts || {}
 
       const filterRootFn = root
         ? (branch) => branch.length > 0 && branch[0][0] === root
         : () => true
 
       const filterTombstoneOrNot = (branch) => {
```
arj03, line 283: I think this function is one of those cases where I would maybe add a comment to each branch to explain why it is correct

arj03 (approved):
Thanks, looks good

PR #58, opened 2021-11-05
Fix findOrCreate & recps

arj03:
`findOrCreate` was not working when encrypting. First commit fails, second fixes.

### DIFF
--- a/api.js
+++ b/api.js
@@ -98,22 +98,20 @@ exports.init = function (sbot, config) {
         sbot.db.add(msgVal, (err, addedMsg) => {
           if (err) return cb(err)
 
-          function hydrate(msg) {
-            const hydratedSubfeed = sbot.metafeeds.query.hydrateFromMsg(
-              msg,
-              metafeed.seed
-            )
-            cb(null, hydratedSubfeed)
-          }
-
           if (encrypted)
-            sbot.db.get(addedMsg.key, (err, msgVal) =>
-              hydrate({
+            sbot.db.get(addedMsg.key, (err, msgVal) => {
+              const msg = {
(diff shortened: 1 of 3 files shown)


test/api.js
```diff
@@ -357,6 +357,50 @@ tape('findAndTombstone and tombstoning branchStream', (t) => {
   })
 })
 
+tape('findOrCreate() recps', (t) => {
+  const boxdir = '/tmp/metafeeds-metafeed-box2'
+  const boxKey = ssbKeys.loadOrCreateSync(path.join(dir, 'secret'))
```
staltz, line 362: Is it on purpose that the path arg uses `dir` not `boxdir`?

arj03: No, that's a bug. Will fix

api.js
```diff
@@ -86,16 +86,34 @@ exports.init = function (sbot, config) {
...
-          )
-          cb(null, hydratedSubfeed)
+
+          function hydrate(msg) {
+            const hydratedSubfeed = sbot.metafeeds.query.hydrateFromMsg(
+              msg,
+              metafeed.seed
+            )
+            cb(null, hydratedSubfeed)
+          }
```
staltz, line 107: Maybe we don't need this function because it's just a wrapper around `hydrateFromMsg` (after all both the wrapped and the wrapper are "hydrating from a message"). My main concern is memory usage and performance, because this function is created/defined for every `create()` call and it doesn't seem really necessary.

So on lines 111 and 116, just call hydrateFromMsg

arj03: Thanks good idea.

staltz (approved):

PR #61, opened 2022-05-26
refactor README.md

mixmix:
## Motivation / Approach

for this : I wanted to read this in a way which was helpful for beginners. Taking from teaching, the idea I applied was "give me just enough detail to get started, then come back around with more detail as required".

So we start out **very** simple, then put more detail in the API section.

## :fire: I also mutated the API to make things clearer

As I was updating this I stumbled across things which felt like they made telling the story _harder_.
Things like:
- API behaves differently for the root feed than for all other feeds. This means we have to write more things
- use of words which conflict with other modules in ecosystem: `feedtype` is actually more like BFE `format`

I've commented those with :fire: so we can discuss.

_Andre I could have split that out into a different PR but I was writing from the point of view of ideal end state. Lets decide what we want then mutate this PR to end up where we want to be_

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,48 @@ SPDX-FileCopyrightText: 2021 Anders Rune Jensen
 SPDX-License-Identifier: CC0-1.0
 -->
 
-# SSB meta feeds
+# ssb-meta-feeds
 
-An implementation of the [ssb meta feed spec] in JS as a secret stack
-plugin.
+An implementation of the [ssb meta feed spec] in JS as a secret stack plugin.
+The core idea is being able to split out content you publish into _subfeeds_,
+which allows for quicker more targetted replication by peers (only get the subfeeds you want).
+
+
+```mermaid
+graph TB
+
+main
+
+classDef default fill:#3f506c,stroke:#3f506c,color:#fff;
+```
+How "classic" scuttlebutt worked - each device has one `main` feed with all messages
+
+```mermaid
+graph TB
+
+root:::root
+
+root-->aboutMe
+root-->contacts
+root-->posts
+root-->main:::legacy
+
+classDef root fill:#8338ec,stroke:#8338ec,color:#fff;
+classDef default fill:#3a86ff,stroke:#3a86ff,color:#fff;
+classDef legacy fill:#3f506c,stroke:#3f506c,color:#fff;
+```
+How scuttlebutt works with meta-feeds - each device now has a `root` feed,
+whose sole responsibility is to announce (point to) sub-feeds that you publish content to.
+
+This means that when you first meet a peer you can replicate their `root` feed
+and (having discovered their subfeeds), replicate just their `aboutMe` and `contacts` feeds
+to get enough info to place them socially. Once you decide you want to follow them you may
+replicate their other subfeeds.
+
+
+_NOTE: The ideal state is that all content is split out into subfeeds.
+To add backwards compatability for devices that have already posted a lot of posts to their
+classic `main` feed, this library will auto-link that main feed in as a "subfeed" of our root._
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


mixmix:
# Rendered README

https://github.com/ssbc/ssb-meta-feeds/blob/04d9b8e59636b35b852fc07d97b1fa5fac0dfdf0/README.md

README.md
```diff
@@ -31,82 +69,134 @@ Add this plugin like this:
...
+)
+const details = {
+  feedpurpose: 'aboutMe',
+  feedformat: 'classic',
+}
+
+sbot.metafeeds.findOrCreate(null, find, details, (err, aboutMeFeed) => {
+  console.log(rootFeed)
+
+  // publish things to aboutMeFeed!
```
mixmix, line 86: Gave the simplest most common use example I could think of.
Showed how to use the feed object thingy to publish something

README.md
```diff
@@ -31,82 +69,134 @@ Add this plugin like this:
...
+  name: 'baba yaga'
+  description: 'lives in a hutt in the forest, swing by sometime!'
+}
+sbot.db.publishAs(aboutMeFeed.keys, content, cb)
+```
+
+
+## API
+
+### `sbot.metafeeds.findOrCreate(cb)`
```
mixmix, line 105: wrote and explicit signature documentation for this special usage to make it clear.
I think this should be moved down in the docs

:fire:
I disliked that the object you got back from this call of this method was different, so I made it isomorphic:
- metafeed: null
- feedpuprpose: 'root'
- feedtype: 'bendybutt-v1',
- metadata: {}

NOTE I think we should change `feedtype` to `format`

README.md
```diff
@@ -31,82 +69,134 @@ Add this plugin like this:
...
+- `metafeed` - the metafeed you are finding/ creating under, can be:
+    - *FeedDetail* object (as returned by `findOrCreate()` or `getRoot()`)
+    - *null* which is short-hand for the `rootFeed` (this will be created if doesn't exist)
+- `find` - method you use to find an existing *FeedDetail*, can be:
+    - *function* of shape `(FeedDetail) => boolean`
+    - *null* - this method will then return an arbitrary subfeed under provided `metafeed`
+- `details` - used to create a new subfeed if a match for an existing one is not found, can be 
+    - *Object*: 
+        - `details.feedpurpose` *String* any string to characterize the purpose of this new subfeed
+        - `details.feedformat` *String* either `'classic'` or `'bendybutt-v1'`
```
mixmix, line 156: :fire: yeah `feedformat` NOT `feedtype`

mixmix: oh yeah 
:fire: `id` instead of `subfeed` - the whole meta / sub thing is way too many concepts for beginners. here you can easily see `id === keys.id`. `FeedDetail.id` should be the unique identifier of the Feed we're considerring ... calling `FeedDatail.subfeed` is more like asking for this things subfeed. Very confusing

I was also tempted to change `metafeed` to `parentfeed` . More happy to leave this though

README.md
```diff
@@ -31,82 +69,134 @@ Add this plugin like this:
...
+  id,
+  // seed
+  // keys
+  metadata
+}
 ```
-{ feedpurpose, feedformat, metafeed, metadata }
-```
+
+NOTE - may include `seed`, `keys` if this is one of your feeds.
```
mixmix, line 199: :fire: is this true?
This current signature really suggests you can use this on other peoples feedIds too....

README.md
```diff
@@ -179,61 +269,31 @@ The `opts` argument can have the following properties:
...
 
-`metafeed` must be a meta feed object `{ seed, keys }` (as returned by
-`getRoot()`).
-
-`visit` must be a function of the shape `({feedpurpose, subfeed, metadata, keys}) => boolean`.
-
-`reason` must be a string to describe why the found feed is being tombstoned.
+Arguments: 
+- `metafeed` *FeedDetail* object (as returned by e.g. `findOrCreate()`, `getRoot()`).
+- `find` *function* of the shape `(FeedDetail) => boolean`.
```
mixmix, line 283: Introduced this type `FeedDetail` (up for changing that name) to make it easier to talk about everywhere less verbosely.
And make it easier to update docs if that thing changes

README.md
```diff
@@ -31,82 +69,134 @@ Add this plugin like this:
...
+const find = (feed) => (
+  feed.feedpurpose === 'aboutMe' &&
+  feed.feedformat === 'classic'
+)
+const details = {
+  feedpurpose: 'aboutMe',
+  feedformat: 'classic',
+}
+
+sbot.metafeeds.findOrCreate(null, find, details, (err, aboutMeFeed) => {
```
staltz, line 83: This first `null` argument may be confusing for beginners, and it's not explained that `null` is actually a short for "root" feed. I think you could restore the old example of first creating a root metafeed and then explicitly passing that root metafeed to this first argument, when creating the subfeed. This way readme readers can understand that we're "finding something" under that root metafeed, and then "creating under that root metafeed" in case it doesn't exist.

README.md
```diff
@@ -31,82 +69,134 @@ Add this plugin like this:
...
-private message to yourself, to allow recovering it in the future.
+Once you have these the *FeedDetail* object `aboutMeFeed` you can publish to that feed:
 
-There can only be one _root_ meta feed, so even if you call `findOrCreate` many
-times, it will not create duplicates, it will just load the meta feed
-`{ seed, keys }`.
+```js
+const content = {
+  type: 'about',
+  about: rootFeed.keys.id,
```
staltz, line 95: `rootFeed` is used here but it wasn't declared anywhere above it. It's good if readme readers can just copy paste the examples and they can actually run.

README.md
```diff
@@ -31,82 +69,134 @@ Add this plugin like this:
...
 ```
 
-This has created a `keys` object for a new subfeed, which you can use to publish
-application-specific messages (such as for a game). The `details` argument
-always needs `feedpurpose` and `feedformat` (supports `classic` for ed25519
-normal SSB feeds, and `bendybutt-v1`).
+Meaning:
+- `metafeed` - the id of the feed this is underneath. As this is the topmost feed, this is empty
+- `feedpurpose` - a human readable ideally unique handle for this feed
+- `feedtype` - the feed format ("classic" or "bendybutt-v1" are current options)
```
staltz, line 128: Where did you find `feedtype`? I grepped ssb-meta-feeds and couldn't find it.

README.md
staltz: I gave feedback about the naming `subfeed` when the meta feeds spec was being written, I think my suggestion was `id` or `feedId`. But it ended up staying as `subfeed`, and it's part of the spec, such that [go-metafeed](https://github.com/ssbc/go-metafeed) also uses `subfeed`. Not worth changing this name, IMO.

README.md
```diff
@@ -179,61 +269,31 @@ The `opts` argument can have the following properties:
...
 or called with an error object on the 1st argument if it failed.
 
+
 ### `sbot.metafeeds.getRoot(cb)`
 
 Looks for the root meta feed declared by your main feed, and returns it (as
 `{ seed, keys}`) via the callback `cb` if it exists.
 
-If it does not exist, this API will **not** create the root meta feed.
+NOTE: If it does not exist, this API will **not** create the root meta feed.
```
staltz, line 295: What's the purpose of "NOTE:"? Doesn't seem like it adds to the description.

README.md
```diff
@@ -179,61 +269,31 @@ The `opts` argument can have the following properties:
...
-possible error, and the 2nd argument is the created feed (which can be either
-the root meta feed `{ seed, keys }` or a sub feed
-`{ feedpurpose, subfeed, keys, metadata }`).
-
-### `sbot.metafeeds.findAndTombstone(metafeed, visit, reason, cb)`
+### `sbot.metafeeds.findAndTombstone(metafeed, find, reason, cb)`
 
 _Looks for the first subfeed of `metafeed` that satisfies the condition in
-`visit` and, if found, tombstones it with the string `reason`.
+`find` and, if found, tombstones it with the string `reason`.
```
staltz, line 276: I prefer the name `visit`, because "find" is the name of the method and "visit" is the name of the argument.

README.md
mixmix: You might be right.
I was trying to think "what is the most common use case of this".
And this is it (I think), but you're right it doesn't aid in learning in the same way.

Will add back in

README.md
mixmix: In this case this example is bad too!
Should not publish about messages to rootFeed..
 Right?

README.md
mixmix: Hmmm, not sure.

mixmix: Yeah this is weird. I remember thinking "this shouldn't be type, it's more like bfe format..."
Then I wrote the wrong one?
Strange

README.md
mixmix: Thanks for context. Will revert given its locked in already

README.md
staltz: Well, it's publishing on the aboutMeFeed, so that's fine.

README.md
mixmix: How about isFeed
I think that's clearer

README.md
mixmix: Happy to remove. I like Note of there's a perhaps expected or important side effect or behavior

README.md
mixmix: god that's so inaccessible!
I'm going with `isFeed` - says what it is on the tin

README.md
mixmix: addressed in https://github.com/ssbc/ssb-meta-feeds/pull/62

mixmix:
@staltz have addressed all comments. thanks for review

Do you want to re-review, or shall I merge?

mixmix:
lol... changes README, tests fail D:

staltz:
> have addressed all comments. thanks for review
> 
> Do you want to re-review, or shall I merge?

@mixmix Let's merge, but I'll make a follow-up PR to recover some of the text you deleted, like under the "Example usage" section, there was quite a good amount of explanation and hand-holding that shouldn't be deleted.

As a general rule that arj and I have been following, the PR author merges only after the PR reviewer has approved with a green checkmark. Any moment before that is considered "don't merge yet".

staltz (approved):

staltz (approved):

PR #62, opened 2022-05-30
make root FeedDetails be same shape

mixmix:
when doing learning I expected the shape of the FeedDetails that come back from methods to be the same.
I noticed the `rootFeed` was lacking (only had seed, keys) and as a beginner this didn't help me understand the tree structure being built

This PR brings that into alignment. I've based it on the README change I made where I presumed this was done


package.json
```diff
@@ -36,14 +36,13 @@
...
-    "ssb-db2-box2": "^0.4.0",
     "ssb-caps": "^1.1.0",
-    "tap-bail": "^1.0.0",
-    "tap-spec": "^5.0.0",
+    "ssb-db2-box2": "^0.4.0",
+    "tap-arc": "^0.3.4",
     "tape": "^5.3.0"
   },
   "scripts": {
-    "test": "tape test/*.js | tap-bail | tap-spec",
```
mixmix, line 46: :hot_pepper: `tap-spec` generates npm audit high level warnings. I want these gone from repos because they add noise which makes it hard to see actual vulnerabilities.

I researched the problem with `tap-spec` and it's fixable, but this module is no longer maintained. I forked it, but it's not ideal. After a bunch of searching I've settled on `tap-arc` which has really nice diff output, "bail" as builtin functionality, and the developers are a company who are very responsive to PRs/ issues.

mixmix:
I'll remove it and add it to the gitignore

I've not used npmrc before...

mixmix:
Fixed this. The commit history is weird.
Should merge-squash

staltz (approved):
Looks good. Feel free to Squash&Merge

PR #63, opened 2022-05-30
feedformat > format, feedpurpose > purpose

mixmix:
Couple of things irked me about `feedformat` + `feedpurpose`:
- not camel case,
- verbose - the context where there are used ... you're already dealing with a feed, it's like making the `key` on a message be called `messagekey` (imo).

This PR:
- :heavy_check_mark: changes `feedformat` for `format`
- :heavy_check_mark: changes `feedpurpose` for `purpose`
- :x: fails tests because test vector signatures ??

Figured I would check  in on this before going further (e.g. it may be this is all already locked in because of an in production use of this)

:fire: BREAKING -  major version change

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,12 +75,12 @@ sbot.metafeeds.findOrCreate((err, rootFeed) => {
   console.log(rootFeed)
 
   const isFeed = (feed) => (
-    feed.feedpurpose === 'aboutMe' &&
-    feed.feedformat === 'classic'
+    feed.purpose === 'aboutMe' &&
+    feed.format === 'classic'
   )
   const details = {
-    feedpurpose: 'aboutMe',
-    feedformat: 'classic',
+    purpose: 'aboutMe',
+    format: 'classic',
   }
   sbot.metafeeds.findOrCreate(null, isFeed, details, (err, aboutMeFeed) => {
     console.log(aboutMeFeed)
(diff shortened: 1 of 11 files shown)


staltz:
Sorry Mix, this can't (easily) happen because it's in the spec: https://github.com/ssbc/ssb-meta-feeds-spec and implemented also in Go: https://github.com/ssbc/go-metafeed/blob/acbfc2812aa5a58a97aa25230a0ccfcda092e5db/metamngmt/encoding_add.go#L18

I remember giving this same feedback during the design phase of the spec, and can't remember the reason why it wasn't changed, but I remember I accepted that it'll stay like this.

We *could* go about changing the spec and the go implementation, but they would have to be PRs that are merged simultaneously. I don't know if you want to fiddle with the Go codebase. 

Anyway, try to get review/feedback from @arj03 and @mycognosist too who worked on this.

> x fails tests because test vector signatures ??

Because the test vector comes from the Go codebase which assumes `feedpurpose` etc.

mixmix:
If it's too hard then let's not.

If the code is not in production yet, then worth considering IMO

mycognosist:
Thanks for the ping.

I am not in favour of merging this. Doesn't seem worthwhile to make a breaking change for a minor naming preference.

PR #64, opened 2022-05-31
recover some explanations on the readme

staltz:
Follow up for #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,12 @@ SPDX-License-Identifier: CC0-1.0
 # ssb-meta-feeds
 
 An implementation of the [ssb metafeed spec] in JS as a secret stack plugin.
-The core idea is being able to split out content you publish into _subfeeds_,
-which allows for quicker more targetted replication by peers (only get the subfeeds you want).
+The core idea is being able to split out content you publish into _subfeeds_.
+This allows for quicker replication by peers, such that you only get the
+subfeeds, thus content types, you are interested in.
 
+Metafeeds are special types of feeds which own or "contain" other feeds (called
+"subfeeds"), by publishing messages which describe the creation of the subfeeds.
 
 ```mermaid
 graph TB
@@ -43,7 +46,6 @@ This means that when you first meet a peer you can replicate their `root` metafe
 and, having discovered their subfeeds, replicate just their `aboutMe` and `contacts` feeds
 to get enough info to place them socially. Once you decide you want to follow them you may
 replicate their other subfeeds.
-
 
 _NOTE: The ideal state is that all content is split out into subfeeds.
 To add backwards compatability for devices that have already posted a lot of posts to their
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,11 +34,13 @@ classDef root fill:#8338ec,stroke:#8338ec,color:#fff;
...
-whose sole responsibility is to announce (point to) sub-feeds that you publish content to.
+How scuttlebutt works with metafeeds - each device now has a `root` metafeed,
+whose sole responsibility is to announce (point to) subfeeds that you publish
+content to. A subfeed can also be a metafeed, which then allows the existence of
+"sub-subfeeds".
 
-This means that when you first meet a peer you can replicate their `root` feed
-and (having discovered their subfeeds), replicate just their `aboutMe` and `contacts` feeds
+This means that when you first meet a peer you can replicate their `root` metafeed
+and, having discovered their subfeeds, replicate just their `aboutMe` and `contacts` feeds
```
mixmix, line 43: Maybe we should emphasize that these subfeeds are just examples of what you might want.

 replicate just the subfeeds you care about e.g. Aboutme...

README.md
```diff
@@ -69,35 +71,59 @@ Add this plugin like this:
...
+  // Find an existing feed
   const isFeed = (feed) => feed.feedpurpose === 'aboutMe'
-  // details for creating a sub-feed feed if it doesn't already exist
+  // Details for creating a subfeed if it doesn't already exist:
   const details = { feedpurpose: 'aboutMe', feedformat: 'classic' }
 
-  sbot.metafeeds.findOrCreate(rootFeed, isFeed, details, (err, aboutMeFeed) => {
-    console.log(aboutMeFeed)
-
+  sbot.metafeeds.findOrCreate(root, isFeed, details, (err, aboutMeFeed) => {
```
mixmix, line 96: I liked rootFeed, makes it clear that it's the same as aboutMeFeed

Not gonna die, just think it makes it clearer than root

README.md
```diff
@@ -69,35 +71,59 @@ Add this plugin like this:
...
     // ...
   })
 })
 ```
 
-Once you have these the *FeedDetail* object `aboutMeFeed` you can publish to that feed:
+The `details` argumentis a *FeedDetails* object and it always requires
+`feedpurpose` (any string) and `feedformat` (either `'classic'` for normal
+ed25519 SSB feeds, or `'bendybutt-v1'` if you want a metafeed). The `isFeed`
+function also takes a *FeedDetails* object as input.
```
mixmix, line 105: Nice!

README.md
```diff
@@ -69,35 +71,59 @@ Add this plugin like this:
...
-    console.log(aboutMeFeed)
-
+  sbot.metafeeds.findOrCreate(root, isFeed, details, (err, aboutMeFeed) => {
     // ...
   })
 })
 ```
 
-Once you have these the *FeedDetail* object `aboutMeFeed` you can publish to that feed:
+The `details` argumentis a *FeedDetails* object and it always requires
```
mixmix, line 102: Typo : argumentis

mixmix (approved):
Only one typo. Rest is opinions. Looks better i agree.

One thing I find a bit wearisome and confusing is what is a metafeed. What's the difference between a feed and a meta feed? 
I think we should say (if we don't already), that a metafeed does pointing to other feeds (and nothing else), while all other application logic is kept in "normal" feeds. 

Then say the root feed is a meta-feed (points to other subfeeds). Then after that just call it the root feed. Or root feed (topmost meta feed).

I think the confusing aspect is that meta means "higher" abstraction to me while root means... kinda lower. It feels like mixed metaphors to me when I first read it. I'm now getting used to it, but we want it very clear for beginners

staltz:
Thanks for the review! I made the typo correction plus some metafeed conceptual explanation.

PR #68, opened 2022-07-26
update to ssb-db2 version 5.0.0

staltz:
The git diff is net negative :) probably because we don't need `getLatest` anymore.


package.json
```diff
@@ -23,29 +23,30 @@
     "pull-cat": "^1.1.11",
     "pull-notify": "^0.1.1",
     "pull-stream": "^3.6.14",
-    "ssb-bendy-butt": "~0.12.5",
```
staltz, line 26: This is now just a feed format that has to be installed as an attachment on ssb-db2, so ssb-meta-feeds doesn't import it directly.

messages.js
```diff
@@ -72,70 +51,40 @@ exports.init = function init(sbot) {
...
-      mfKeys,
-      previous,
-      feedpurpose,
-      seed,
-      feedformat,
-      metadata
-    ) {
-      if (feedformat !== 'classic' && feedformat !== 'bendybutt-v1') {
-        throw new Error('Unknown feed format: ' + feedformat)
+    optsForAddDerived(mfKeys, feedpurpose, seed, feedFormat, metadata) {
```
arj03, line 60: Couldn't derived also be buttwoo now?

arj03: Totally fine that this one of just a 1 to 1 of existing functionality

staltz: Yeah, let's do a PR for buttwoo too, so we can add tests for it

arj03 (approved):
Great to see how the new simplifies this module

PR #69, opened 2022-09-14
Auto shard

mixmix:
~~This is WIP~~

**Problem**: the meta-feed spec describes where feeds should be located in nibble-shards. We want to save users from having to figure this out, so we need a nice methods for this

I wasn't sure what we want the final API of this module should be. I'd like to discuss this, but for the moment I've just put this method at a placeholder location `ssb.metafeeds.v1.findOrCreate(details, cb)`

I'll raise question I've encountered as comments in code


README.md
```diff
@@ -130,6 +130,41 @@ sbot.metafeeds.findOrCreate((err, root) => {
 
 ## API
 
+### `sbot.metafeeds.v1.findOrCreate(details, cb)`
```
mixmix, line 133: I expect this to be one of the most-used functions, so it feels weird nesting it under v1 like this.
To what degree do we want the final API to lower level methods, to help guide people to use "sensible defaults" like this one. Is this a sensible default or just following the spec? (we still want advanced uses to be able to make arbitrary nesting I guess right?)

README.md
```diff
@@ -130,6 +130,41 @@ sbot.metafeeds.findOrCreate((err, root) => {
...
+
+Looks for the first subfeed of `metafeed` that matches `details`, or creates on which matches these.
+This creates feeds following the [v1 tree structure](https://github.com/ssbc/ssb-meta-feeds-spec#v1).
+
+Arguments:
+- `details` *Object* where 
+    - `details.feedpurpose` *String* any string to characterize the purpose of this new subfeed
+    - `details.feedformat` *String* (optional)
+        - either `'classic'` or `'bendybutt-v1'`
+        - default: `'classic'`
```
mixmix, line 143: should we have a default format?

README.md
mixmix: keeping with the theme of high-level sensible default, i've dropped the `isFeed` function from this signature. hmmm

api.js
```diff
@@ -214,6 +216,38 @@ exports.init = function (sbot, config) {
...
+      if (err) return cb(err)
+
+      const v1Details = { feedpurpose: 'v1', feedformat: BB1 }
+      const v1Visit = (feed) => feed.feedpurpose === v1Details.feedpurpose
+
+      findOrCreate(rootFeed, v1Visit, v1Details, (err, v1Feed) => {
+        if (err) return cb(err)
+
+        const shardDetails = {
+          feedpurpose: pickShard(rootFeed.keys.id, details.feedpurpose),
```
mixmix, line 232: :hot_pepper: the spec says

> When adding a new application-specific subfeed to the tree, we need to determine the parent shard based on a "name", which is any UTF-8 string that the application can choose freely, but it is RECOMMENDED that this string be unique to the use case

However I've just used the `feedpurpose` of the leaf subfeed. Is this wrong? Should the spec say you should use the `feedpurpose`, or should we have a new field on details like `feedidkeykey` (id and key are already taken :cry:)

pick-shard.js
```diff
@@ -0,0 +1,15 @@
...
+const bfe = require('ssb-bfe')
+const crypto = require('crypto')
+
+module.exports = function pickShard(rootFeedId, idString) {
+  const buf = Buffer.concat([bfe.encode(rootFeedId), bfe.encode(idString)])
+
+  const hash = crypto.createHash('sha256')
+  hash.update(buf)
+
+  return hash.digest('hex')[0]
```
mixmix, line 14: :fire: this always returns strings e.g. `'a'` or '`7'`
This is actually good because the `feedpurpose` for the shard MUST be a string (according to this module API anyway). Should we change the spec to clarify that it's the first hex character **as a string** that defines the shard?

test/api.js
```diff
@@ -17,14 +17,15 @@ const mainKey = ssbKeys.loadOrCreateSync(path.join(dir, 'secret'))
...
-    keys: mainKey,
-    path: dir,
-  })
+function createSbot(config = { keys: mainKey, path: dir }) {
+  return SecretStack({ appKey: caps.shs })
+    .use(require('ssb-db2'))
+    .use(require('ssb-bendy-butt'))
+    .use(require('../'))
+    .call(null, config)
+}
```
mixmix, line 26: why are we not using `scuttle-testbot`. It hurts me seeing this code and rimraf and key creation repeated so much.
If there's a good reason it's fine, but I tend to like to do something like this, which has a great simple API, and means one place to maintain test server

```js
// testbot.js
const Server = require('scuttle-testbot')

module.exports = function (opts = {}) {
  // opts = {
  //   startUnclean: Boolean,    - whether to rimraf
  //   keys: SecretKeys
  // }

  const stack = Server
    .use(require('ssb-backlinks'))
    .use(require('ssb-query'))
    .use(require('../'))

  return stack({
    db2: true,
    ...opts
  })
}
```

test/api.js
```diff
@@ -402,6 +387,42 @@ tape('findOrCreate() recps', (t) => {
...
+  }
+  sbot.metafeeds.v1.findOrCreate(details, (err, feed) => {
+    if (err) throw err
+    t.equal(feed.feedpurpose, details.feedpurpose, 'creates feed')
+
+    sbot.metafeeds.v1.findOrCreate(details, (_, feed2) => {
+      t.deepEqual(feed, feed2, 'finds feed')
+
+      pull(
+        sbot.metafeeds.branchStream({ root: null, old: true, live: false }),
```
mixmix, line 408: :fire: I may not have understood, but this is returning `[]` while the tests before it pass.... feeds are being created.

HALP!

mixmix:
TODO : export some test-vectors for sharding (to make it easier for Go / rust butts to align implementation)

README.md
staltz: Yes classic seems fine :)

pick-shard.js
staltz: Yes, good spotting.

It could be something like

```diff
-The `content` on the root's message for the shard feed MUST have the nibble encoded as hexadecimal in the `feedpurpose` field of the `metafeed/add/derived` message.
+The `content` on the root's message for the shard feed MUST have the nibble expressed in hexadecimal and encoded as a string in the `feedpurpose` field of the `metafeed/add/derived` message.
```

staltz:
> I wasn't sure what we want the final API of this module should be. I'd like to discuss this

I also don't know what the final API should be. Might want to check @arj03 for ideas too.

One idea is to provide only a few APIs: `findOrCreate` (which is your new v1 findOrCreate) and `getOrCreateRoot` (name can be bikeshedded) and `branchStream`, and then tuck away all the other APIs under a `lowlevel` namespace or something like that. With the intention that the README only has the high level APIs, and we keep the low level APIs undocumented, sort of like a back stage access, that we can officialize if they seem important enough.

README.md
staltz: I'm not sure about the removal of `isFeed`. I'm warming up to removing it, but part of me intuitively feels we might need it.

api.js
staltz: Ah, I see, the spec doesn't actually say what the `name` should be. I guess the `name` should be predictable for peers who want to replicate the leaf, because they need to calculate the shard nibble themselves.

I can't remember why we didn't just say that `name = feedpurpose` when drafting the spec. Do you remember @arj03 ?

test/api.js
staltz: I guess it's because scuttle-testbot wouldn't vastly reduce lines of code for us. Honestly it's just about the rimraf, which is not too bad. The key creation part is something we want to control. scuttle-testbot generates random keys and sometimes we want to control the seed so that the key is always deterministic for the tests.

test/api.js
staltz: I'm not sure I understood, can you clarify?

staltz:
You know what, we would *vastly* simplify our lives if we dropped the possibility that a content leaf is a metafeed too. It would make implementing these changes in ssb-meta-feeds easier, and it would vastly simplify our upcoming changes to ssb-replication-scheduler.

To clarify, the metafeeds tree structure v1 spec would allow this (focus on `chess` and `tictactoe`):

```mermaid
graph TB;
  root --> v1
  v1 --> 7 & b & f
  7 --> games
  games --> chess & tictactoe
  f --> posts
  b --> abouts
```

What if we would change the spec so that tree structure v1 is rigid, like this:


```mermaid
graph TB;
  root --> v1
  v1 --> 7 & b & f
  7 --> chess & tictactoe
  f --> posts
  b --> abouts
```

We can always just make tree structure v2 if we find a compelling case for that.

staltz:
@mixmix Another thing: remove the package-lock.json that was added

test/api.js
mixmix: you can create your own keys IF required with scuttle-testbot. (generally not needed apart from persistence/ restart tests which a rare).
the module also takes care of making sure your `path` never collides (unless on the other hand you want it to, in which case you set it. Again rare)

mixmix: MEH

mixmix:
I like your suggestion about "flat" leaves. Then we can hide almost all the bullshit...... OH I remember I was keen to have the `unique name !== purpose` because I argued for "clustering" of feeds

e.g. I might want to have `chess/games` and `chess/meetups` be in the same shard, because:
- EITHER I'm a chess player, in which case i want
    - only the games feed
    - and MAYBE I want to upgrade to have the meetups for in person gathering of chess players
- OR I'm not a chess player in which case I want NEITHER feed.

So we could make `unique name = purpose.split('/')[0]`

I dunno....

mixmix:
I like your suggestion about "flat" leaves. Then we can hide almost all the bullshit...... API which lets you make meta sub-feeds etc. and you really have to go looking to find those API

staltz:
> flat leaves

Yeah, this is going to require a PR to ssb-meta-feeds-spec.

lookup.js
```diff
@@ -195,6 +195,7 @@ exports.init = function (sbot, config) {
   }
 
   function branchStreamOld(rootMetafeedId) {
+    // WARNING this method needs loadState to be called?
```
staltz, line 198: This is already handled when calling `branchStream()`. This comment can safely be removed.

mixmix: I just say your change to add defer stuff! nice!

test/api.test.js
```diff
@@ -0,0 +1,374 @@
...
+        // TODO it would be nice for testing that we could deterministically know the shard
+        // but I don't know how to fix the "seed" that the root feed is derived from
+        resolve()
+      })
+    )
+  })
+
+  sbot.close()
+  t.end()
+})
```
mixmix, line 374: TODO - consider splitting out the "advanced" methods into own tests
Perhaps extract high level sugar methods into own tests - these are more integration tests, but nice to give them space

api.js
```diff
@@ -211,10 +245,24 @@ exports.init = function (sbot, config) {
...
+      findById,
+
+      // TODO add API for whether to encrypt the announce message for a subfeed
+      // for sympathetic replication?
+      //
+      // ssb.db.create // allows
+      //
+      // QUESTION: should the groupKey be used for the determination of the location of the group subfeed (ideally don't use the same key for two different things right?)
+
+      // TODO findOrCreate thing for friends helsinki subfeed?
```
mixmix, line 264: different PR, create issue

api.js
```diff
@@ -211,10 +245,24 @@ exports.init = function (sbot, config) {
...
+    findOrCreate: v1FindOrCreate,
+
+    advanced: {
+      getRoot,
+      findOrCreate,
+      findAndTombstone,
+      findById,
+
+      // TODO add API for whether to encrypt the announce message for a subfeed
+      // for sympathetic replication?
```
mixmix, line 258: scope out the API change, make sure this API is flexible-foundation, then do that in different PR

api.js
```diff
@@ -211,10 +245,24 @@ exports.init = function (sbot, config) {
...
+      //
+      // ssb.db.create // allows
+      //
+      // QUESTION: should the groupKey be used for the determination of the location of the group subfeed (ideally don't use the same key for two different things right?)
```
mixmix, line 262: TODO - update spec to pin this down

README.md
staltz: I can't remember what we spoke when we were designing this in the NGI pointer grant.

But think of it from the perspective of a new developer who has little context for all these tools, and they see the field `feedpurpose`, I think it's reasonable that they may think "I can have two feeds with the same purpose".

Since in the metafeed tree there are going to be subfeeds generated by different app developers, it's reasonable to assume that these developers may be stepping on each other's toes, and that's why they might need more granularity, with `visit`

api.js
staltz: By the way, I don't think TODOs should be inside code, we should aim for converting these to issues before we merge this PR.

api.js
staltz: https://github.com/ssbc/ssb-meta-feeds/issues/79

api.js
staltz: https://github.com/ssbc/ssb-meta-feeds/issues/78

staltz:
I converted this PR to "draft" because it needs to be rebased on `master`.

README.md
```diff
@@ -21,7 +21,7 @@ main
```
mixmix, line 21: updated the API to match our new  high-level + advanced split.
The examples are much simpler now.
I moved the advanced APIs to `README_ADVANCED.md` to keep this landing page simpler/ less confusing. There's a link out to it.

NOTE : considered pulling tombstone into "high level" space, but then realised we should agree on the API of that before doing that (would be good for it to align with the other high level `findOrCreate`) = seperate PR

api.js
```diff
@@ -3,9 +3,23 @@
...
 const alwaysTrue = () => true
+const BB1 = 'bendybutt-v1'
+const v1Details = { feedpurpose: 'v1', feedformat: BB1 }
+const v1Visit = detailsToVisit(v1Details)
+function detailsToVisit(details) {
+  return (feed) => {
+    return (
+      feed.feedpurpose === details.feedpurpose &&
+      feed.feedformat === details.feedformat &&
+      deepEqual(feed.metadata, details.metadata || {})
```
mixmix, line 19: I think this is how we want to deep-equal check if a feed matches our details

api.js
```diff
@@ -3,9 +3,23 @@
...
 const run = require('promisify-tuple')
+const deepEqual = require('fast-deep-equal')
 const debug = require('debug')('ssb:meta-feeds')
+const pickShard = require('./pick-shard')
 
 const alwaysTrue = () => true
+const BB1 = 'bendybutt-v1'
+const v1Details = { feedpurpose: 'v1', feedformat: BB1 }
+const v1Visit = detailsToVisit(v1Details)
+function detailsToVisit(details) {
```
staltz, line 14:
```suggestion

function detailsToVisit(details) {
```

test/api.test.js
```diff
@@ -410,3 +410,71 @@ test('findOrCreate() recps', (t) => {
...
+
+    sbot.metafeeds.findOrCreate(details, (err, chessF2) => {
+      if (err) throw err
+      t.deepEqual(chessF, chessF2, 'finds feed')
+
+      pull(
+        sbot.metafeeds.branchStream({ root: null, old: true, live: false }),
+        pull.collect((err, branches) => {
+          if (err) throw err
+          // console.log(branches.map(branch => branch.map(f => f[1] && f[1].feedpurpose)))
```
staltz, line 436:
```suggestion
```

test/api.test.js
```diff
@@ -410,3 +410,70 @@ test('findOrCreate() recps', (t) => {
...
+          // root, v1, shard, chess (AND MAIN)
+
+          const purposePath = branches
+            .pop()
+            .map((f) => f[1] && f[1].feedpurpose)
+          t.deepEqual(purposePath, [null, 'v1', purposePath[2], 'chess'])
+          // TODO it would be nice for testing that we could deterministically know the shard
+          // but I don't know how to fix the "seed" that the root feed is derived from
+
+          sbot.close()
```
staltz, line 447:
```suggestion
          sbot.close(true, t.end)
```

test/api.test.js
```diff
@@ -410,3 +410,70 @@ test('findOrCreate() recps', (t) => {
...
+          t.end()
```
staltz, line 448:
```suggestion
```

test/api.test.js
```diff
@@ -410,3 +410,70 @@ test('findOrCreate() recps', (t) => {
...
+    metadata: {
+      recps: [sbot.id],
+    },
+  }
+
+  sbot.metafeeds.findOrCreate(details, (err, chessF) => {
+    if (err) throw err
+
+    t.deepEqual(chessF.metadata.recps, [sbot.id], 'creates encrypted subfee')
+    sbot.close()
```
staltz, line 476:
```suggestion
  sbot.close(true, t.end)
``

File: test/api.test.js [hunk 19]
```diff
@@ -410,3 +410,70 @@ test('findOrCreate() recps', (t) => {
...
+
+  const ownKey = Buffer.from(
+    '30720d8f9cbf37f6d7062826f6decac93e308060a8aaaa77e6a4747f40ee1a76',
+    'hex'
+  )
+  sbot.box2.setOwnDMKey(ownKey)
+
+  const details = {
+    feedpurpose: 'chess',
+    // feedformat: 'classic', optional
```
staltz, line 466:
```suggestion
```

staltz (approved):
Roger to go

PR #72, opened 2022-09-16
remove unused APIs ensureLoaded, findByIdSync, loadState

staltz:
We only used `loadState`, `ensureLoaded`, and `findByIdSync` in ssb-ebt, but after [this refactor](https://github.com/ssbc/ssb-ebt/pull/71/files) (merged and published), we don't need these anymore.

Should make it overall easier to work on this module if it has less code.

### DIFF
--- a/feeds-lookup.js
+++ b/feeds-lookup.js
@@ -6,6 +6,7 @@ const { seekKey } = require('bipf')
 const pull = require('pull-stream')
 const cat = require('pull-cat')
 const Notify = require('pull-notify')
+const Defer = require('pull-defer')
 const DeferredPromise = require('p-defer')
 const {
   where,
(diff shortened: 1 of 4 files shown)


staltz:
Also @Powersource

arj03 (approved):
Looks good. Did a side by side view of master and this branch (diff is not easy to read in feeds-lookup) and there is almost no changes in those 2 functions.

feeds-lookup.js
```diff
@@ -230,94 +229,83 @@ exports.init = function (sbot, config) {
...
-    },
+  function branchStream(opts) {
+    if (!loadStateRequested) {
+      loadState()
+      const deferred = Defer.source()
+      stateLoadedP.promise.then(() => {
+        deferred.resolve(branchStream(opts))
+      })
+      return deferred
+    }
```
staltz, line 270: This whole if() is an important change in `branchStream`, it wasn't there before. All the rest should be the same.

This if() guarantees that the state is loaded before the stream can be consumed.

arj03: yep :)

PR #73, opened 2022-09-20
make tests atomic, add persistence tests

mixmix:
This PR:
- shifts tests to be atomic (can be run one at a time if required)
- adds persistence tests


test/api.test.js
```diff
@@ -0,0 +1,331 @@
...
+  )
+
+  while (todo > 0) await p(setTimeout)(100)
+
+  console.log('(persistence)')
+  await p(sbot.close)()
+  sbot = Testbot({ path, rimraf: false })
+
+  todo = 3
+  testOldStreams(sbot)
```
mixmix, line 298: these test fail after sbot restart.... not sure why. Output suggested the branches have not been loaded correctly.
That or I'm expecting different behaviour than you!

image

staltz: Why was this test rewritten with promises? It seems like you could have just used the new `Testbot()` helper and kept all the rest the same as before. With less things changing, you're more certain as to why things are breaking.

So my suggestion is to bring back the test as it was before, but just using Testbot() this time.

staltz: https://github.com/ssbc/ssb-meta-feeds/pull/74

PR #75, opened 2022-09-20
add persistence tests to api

mixmix:
Your cb tests dropped my persistence testing work.
This PR adds it back in

PR #76, opened 2022-09-21
testbot + group non-atomic tests

mixmix:
I noticed we weren't using the testbot everywhere. Changed that to main maintenance easier.
I also considered making all test atomic so the repo was consistent, but the remaining tests were too fidey to do this with - instead I just modified the structure to make it very clear that tests depend on one another
- left a comment
- used `t.test` to provide a clear scope
    - this is needed so that you can have `t.only` work because starting sbots in the global scope means `t.only` was never exiting :angry:


test/messages.test.js
```diff
@@ -19,199 +16,196 @@ const metafeedKeys = keys.deriveFeedKeyFromSeed(
...
+test('messages', (t) => {
+  const sbot = Testbot({ keys: mainKey })
+  const {
+    db,
+    metafeeds: { messages },
+  } = sbot
 
-let tombstoneMsg
+  let addMsg
+  // NOTE - these subtests are not "atomic" - db state is shared between them
```
mixmix, line 29: This diff is annoying. I didn't touch the tests at all other then indenting them, and changing to `t.test`

Note the shared variables and sbot are now "inside" a test scope (not global in the file)

staltz:
I'm confused with what happened to all these PRs. I approved my PR-of-your-PR, but apparently you had changed its base to become a PR-to-master, and now we have the commits in master branch:

- [make tests atomic, add persistence tests](https://github.com/ssbc/ssb-meta-feeds/commit/91099c9b82b0db733ac9a6f3e4415036f245c3d0)
- [revert promise tests to cb tests](https://github.com/ssbc/ssb-meta-feeds/commit/0e7a624fce8f9d42bc7a3ecfb630639b7e8e3bb1)

But then **this** PR also has those two commits, plus a git conflict. Also PR #75 has a git conflict.

What's going on??

PR #77, opened 2022-09-21
bug: close metadata edgecase

mixmix:
While reading the code, I noticed a nasty edge case you could fall into with metadata -  because of how it was merged into bendy butt messages, you could over-write the type and purpose of a metafeed announce message accidentally.

I considered two options:
1. (simple) bar some reserved words from use in metadata
2. store metadata under a dedicated field in message

I went with (2) thinking it would be simple, but the tests turned up a counter edge case with `metadata.recps` being used for announce messages... so ... yes.


messages.js
```diff
@@ -34,7 +34,14 @@ exports.init = function init(sbot) {
 
     if (nonce) content.nonce = nonce
 
-    if (metadata) Object.assign(content, metadata)
```
mixmix, line 37: This was the naughty line

test/api.test.js
```diff
@@ -454,6 +454,37 @@ test('findOrCreate', (t) => {
...
+  const sbot = Testbot()
+
+  const details = {
+    feedpurpose: 'chess',
+    // feedformat: 'classic', optional
+    metadata: {
+      type: 'dogs',
+    },
+  }
+
```
mixmix, line 467: here's the test. Without the changes to the code the latest branch you get from the stream is the "shard" (as the message published that was type `metafeed/add/derived` gets replaced by type `dogs` :dog:  Woof

lookup.js
```diff
@@ -87,20 +87,7 @@ exports.init = function (sbot, config) {
     details.feedformat = validate.detectFeedFormat(content.subfeed)
     details.feedpurpose = content.feedpurpose || details.feedpurpose
     details.metafeed = content.metafeed || details.metafeed
-    details.metadata = {} || details.metafeed
```
mixmix, line 90: this look like another bug 

```diff
- {} || details.metafeed
+ {} || details.metadata
```

please check, I don't know what I'm doing

staltz:
Hmm, before I delve into the details here, I think what you're proposing is a breaking change, not only to ssb-meta-feeds, but also to the "spec" (the implicit spec, i.e. the agreement between peers. We should make the implicit explicit, though).

Like, you may be creating an incompatibility with go-metafeed: https://github.com/ssbc/go-metafeed/search?q=metadata and I'd prefer some other solution which doesn't cause incompat or breaking change.

Also, because @arj03 was involved designing him, let's ask his review.

mixmix:
@arj03 can you please review? (see the initial post for context)

test/api.test.js
arj03: I would rather that this gives an error. You are not supposed to use any reserved words in metadata.

lookup.js
arj03: I don't think this is correct. The metadata is added to the message content as a flat structure (as keys). Not as an object, this confusing part here is maybe that you pass in an object to the findOrCreate.

messages.js
arj03: Yeah, if we check for nuaghty keys instead we should be good.

arj03:
Oh sorry I never saw this. The spec as it is currently (it could be worded a bit better and have better examples, especially for metadata!), but everything in the content is metadata: https://github.com/ssbc/ssb-meta-feeds-spec#example-of-a-metafeed. So I would rather go for option 1 in you description.

mixmix:
Replacing with https://github.com/ssbc/ssb-meta-feeds/pull/87

PR #81, opened 2022-09-27
change branchStream to return consistent shaped results

mixmix:
I changed other APIs such that the rootFeed accessors also returns the same shape details

This PR does the same thing for `branchStream`. Just means less annoying edge case handling when consuming.
The problem I hit was going `branch.map(details => details.feedpurpose)` and getting some stupid error because `details = null` for the root feed. sigh.


staltz (approved):
Great improvement

PR #84, opened 2022-09-28
split tests by api method

mixmix:
No edits, just moved files + functions around.
The intention is to make it easier to track test coverage by having all tests for each method in one place.

Problem we saw was that `branchStream` had a test missing (PR to follow this one), but this was missed as `api.test.js` was so big


staltz (approved):
Much appreciated

staltz:
@mixmix Press the green

PR #85, opened 2022-09-28
Branch stream private

mixmix:
Arj pointed out `branchStream` does not handle private meta-feed announces. This is important because we use/ will use this to inform which feeds we should be replicating (wait is that true, does branchStream stream other peoples feed trees?).

Anyway, this API should work. I have
1. written a failing test
2. started trying to get it passing. Problems:
    - unclear what encryption is being used (is it DMOther style as if `recps = [sbot.id]` and publishing on `rootMF` then that's an "other"....)
    - unclear why `isDecrypted` operator isn't working (no errors, just no decryption, did I use it right?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,12 +126,12 @@ Arguments:
     - `details.feedformat` *String* (optional)
         - either `'classic'` or `'bendybutt-v1'`
         - default: `'classic'`
-    - `details.encrption` *Object*
     - `details.recps` *Array* (optional)
        - A collection of "recipients" (GroupId, FeedId, ...) to encrypt the announcement messages to 
-    - `details.encrptionFormat` *String* (optional)
+    - `details.encryptionFormat` *String* (optional)
        - specifies which encrption format to use (you will need an encrption plugin installed e.g. `ssb-box2` installed)
        - default: `'box2'`
+    - `details.metadata` *Object* (optional) - for containing other data
 
 - `cb` *function* delivers the response, has signature `(err, FeedDetails)`, where FeedDetails is
     ```js
@@ -161,6 +161,7 @@ Meaning:
 - `feedformat` - the feed format ("classic" or "bendybutt-v1" are current options)
 - `seed` - the data from which is use to derive the `keys` and `id` of this feed.
 - `keys` - cryptographic keys used for signing messages published by this feed (see [ssb-keys])
+- `recps` - an Array of recipients who the meta-feed announcement was encrpted to
 - `metadata` - object containing additional data
 
 NOTES:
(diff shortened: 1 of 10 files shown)


mixmix:
@staltz can you look at the content of this and advise any way to progress this? Was semi-sick today so didn't look into this today but would like to today

lookup.js
```diff
@@ -144,7 +144,7 @@ exports.init = function (sbot, config) {
 
     pull(
       sbot.db.query(
-        where(and(authorIsBendyButtV1(), isPublic())),
+        where(and(authorIsBendyButtV1(), isDecrypted())),
```
staltz, line 147: @mixmix I think we can just remove `isPublic()` without adding `isDecrypted()` (context: we used to have "isPrivate()" but we split that into two operators: `isEncrypted()` and `isDecrypted()` and then we re-added `isPrivate()` as an alias for `or(isEncrypted(), isDecrypted())`)

If that doesn't work, then you can try explicitly:

```js
and(authorIsBendyButtV1(), or(isPublic(), isDecrypted()))
```

staltz:
> unclear why isDecrypted operator isn't working (no errors, just no decryption, did I use it right?)

Can you clarify what you mean by "isn't working"? (e.g. expected versus actual)

> unclear what encryption is being used (is it DMOther style as if recps = [sbot.id] and publishing on rootMF then that's an "other"....)

I think we need to first create a decent API or improvement to `findOrCreate` which clearly allows specifying encryption. Maybe it should resemble `ssb.db.create`'s opts, like `opts.encryptionFormat` and `opts.recps`.

mixmix:
Picking up the API change for encryption here: https://github.com/ssbc/ssb-meta-feeds/pull/88

mixmix:
This is on hold while the API for encrypted metafeed stuff is pinned down

README.md
```diff
@@ -220,7 +220,7 @@ and
 
 ```js
 [
-  [rootMetafeedId, null],
+  [rootMetafeedId, rootDetails],
```
mixmix, line 223: changed this API to match the `rootDetails` behaviour with `findAndCreate` (i.e. give back actual details).
Needed this for writing simpler tests (no `null` edge cases, can always ask the purpose)

lookup.js
```diff
@@ -106,7 +115,10 @@ exports.init = function (sbot, config) {
 
     // Update roots
     if (!detailsLookup.has(metafeed)) {
-      detailsLookup.set(metafeed, null)
+      // !detailsLookup.has(metafeed) !== "metafeed is a root)
+      // this assumes that we have replicated the metafeed tree in the order
+      // that it was created.
```
mixmix, line 120: :fire: spotted a potential edge case that could cause us pain later.
Documented in code. Should this be an issue?

README.md
mixmix: ok making this another PR

mixmix: MOVED :heart:

PR #87, opened 2022-10-03
prevent use of protected metadata fields

mixmix:
Does this one simple thing.

Supercedes: https://github.com/ssbc/ssb-meta-feeds/pull/77

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -1,0 +1,18 @@
+const BB1 = 'bendybutt-v1'
+const v1Details = { feedpurpose: 'v1', feedformat: BB1 }
+const NOT_METADATA = new Set([
+  'type',
+  'metafeed',
+  'feedpurpose',
+  'subfeed',
+  'tangles',
+  'reason',
+  'nonce',
+  'recps',
+])
+
+module.exports = {
+  BB1,
+  v1Details,
+  NOT_METADATA,
+}
--- a/constants.js
+++ b/constants.js
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 2022 Mix Irving
+//
+// SPDX-License-Identifier: LGPL-3.0-only
+
 const BB1 = 'bendybutt-v1'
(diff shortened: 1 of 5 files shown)


lookup.js
```diff
@@ -88,16 +89,7 @@ exports.init = function (sbot, config) {
...
-      'metafeed',
-      'feedpurpose',
-      'type',
-      'tangles',
-      'reason',
-      'subfeed',
-      'nonce',
-    ]
-    const keys = Object.keys(content).filter((k) => !NOT_METADATA.includes(k))
+    const keys = Object.keys(content).filter((k) => !NOT_METADATA.has(k))
```
arj03, line 92: why not includes here? It seems to me that has in not an options on array (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) and thus operates on object that just happens to work? Or am I missing something?

test/api/find-or-create.test.js
```diff
@@ -43,27 +43,3 @@ test('findOrCreate', (t) => {
     })
   })
 })
-
-test('findOrCreate (metadata.recps)', (t) => {
```
arj03, line 47: just to be sure I'm reading this right, we don't have any tests for encrypted metafeeds after this commit? AH yes, this is now https://github.com/ssbc/ssb-meta-feeds/pull/88. Okay makes sense 👍

constants.js
```diff
@@ -0,0 +1,22 @@
+// SPDX-FileCopyrightText: 2022 Mix Irving
+//
+// SPDX-License-Identifier: LGPL-3.0-only
+
+const BB1 = 'bendybutt-v1'
+const v1Details = { feedpurpose: 'v1', feedformat: BB1 }
+const NOT_METADATA = new Set([
```
mixmix, line 7: Set

lookup.js
arj03: Ah, did not see that. Perfect

arj03 (approved):
Nice refactor, thanks

PR #88, opened 2022-10-03
encrypted metafeed API

mixmix:
Cases this has to cover:
1. `recp = sbot.id`
   - this is technically an "own_secret" case, but we will hit bumps because ssb-box2 doesn't know about meta-feeds, and "own" now being many keys
       - on encrption, I _think_ box2 will see this as own-encrypt?
       - on decrypt, box2 will almost certainly NOT try own keys

2. `recps = groupId`
   - this case should be fine... as we try group keys on all our messages

3. `recp = someFeedId`
   - this is the DM case, so the shared key should be derived from shardFeedId (author) + someFeedId
   - this should "just work"

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,8 +126,13 @@ Arguments:
     - `details.feedformat` *String* (optional)
         - either `'classic'` or `'bendybutt-v1'`
         - default: `'classic'`
-    - `details.metadata` *Object* (optional) - for containing other data
-        - if `details.metadata.recps` is used, the subfeed announcement will be encrypted
+    - `details.encrption` *Object*
+    - `details.recps` *Array* (optional)
+       - A collection of "recipients" (GroupId, FeedId, ...) to encrypt the announcement messages to 
+    - `details.encrptionFormat` *String* (optional)
+       - specifies which encrption format to use (you will need an encrption plugin installed e.g. `ssb-box2` installed)
+       - default: `'box2'`
+
 - `cb` *function* delivers the response, has signature `(err, FeedDetails)`, where FeedDetails is
     ```js
     {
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -126,8 +126,13 @@ Arguments:
     - `details.feedformat` *String* (optional)
         - either `'classic'` or `'bendybutt-v1'`
         - default: `'classic'`
-    - `details.metadata` *Object* (optional) - for containing other data
-        - if `details.metadata.recps` is used, the subfeed announcement will be encrypted
+    - `details.encrption` *Object*
```
staltz, line 129:
```suggestion
    - `details.metadata` *Object* (optional) - for containing other data
```

I guess?

README.md
```diff
@@ -126,8 +126,13 @@ Arguments:
...
+    - `details.recps` *Array* (optional)
+       - A collection of "recipients" (GroupId, FeedId, ...) to encrypt the announcement messages to 
+    - `details.encrptionFormat` *String* (optional)
```
staltz, line 132:
```suggestion
    - `details.encryptionFormat` *String* (optional)
```

staltz (changes requested):

staltz:
@mixmix Mind deleting the branch too? (in case you're sure it's not going to be used)

mixmix:
Uh.... what? don't know why this was closed. Accident?
How to reopen. Sigh

mixmix:
Oh, the branch it was being merged into was merged and deleted. Sigh

mixmix:
see https://github.com/ssbc/ssb-meta-feeds/pull/88

PR #89, opened 2022-10-03
encrypted metafeed API (AGAIN)

mixmix:
Cases this has to cover:
1. :x: `recp = sbot.id`
   - this is technically an "own_secret" case, but we will hit bumps because ssb-box2 doesn't know about meta-feeds, and "own" now being many keys
       - on encrption, I _think_ box2 will see this as own-encrypt?
       - on decrypt, box2 will almost certainly NOT try own keys

2. :heavy_check_mark:  `recps = groupId`
   - this case should be fine... as we try group keys on all our messages

3. :x: `recp = someFeedId`
   - this is the DM case, so the shared key should be derived from shardFeedId (author) + someFeedId
   - this should "just work"

4. :x: `recp = POBoxId`

:x: = this PR doesn't handle, and has test cases showing they explicitly callback with an error

### DIFF
--- a/api.js
+++ b/api.js
@@ -4,6 +4,7 @@
 
 const run = require('promisify-tuple')
 const deepEqual = require('fast-deep-equal')
+const { isCloakedMsgId } = require('ssb-ref')
 const debug = require('debug')('ssb:meta-feeds')
 const pickShard = require('./pick-shard')
 const { BB1, v1Details, NOT_METADATA } = require('./constants')
@@ -16,19 +17,20 @@ function detailsToVisit(details) {
     return (
       feed.feedpurpose === details.feedpurpose &&
       feed.feedformat === details.feedformat &&
-      deepEqual(feed.metadata, details.metadata || {})
+      (details.metadata ? deepEqual(feed.metadata, details.metadata) : true) &&
+      (details.recps ? deepEqual(feed.recps, details.recps) : true)
     )
   }
 }
 
 exports.init = function (sbot, config) {
   function filter(metafeed, visit, maybeCB) {
+    const cb = maybeCB
     if (!metafeed) {
       cb(new Error('expected metafeed argument'))
     } else if (typeof metafeed === 'function') {
       cb(new Error('expected metafeed argument and visit argument'))
     } else {
-      const cb = maybeCB
       sbot.metafeeds.query.hydrate(
         metafeed.keys.id,
         metafeed.seed,
@@ -43,12 +45,12 @@ exports.init = function (sbot, config) {
   }
 
   function find(metafeed, visit, maybeCB) {
+    const cb = maybeCB
     if (!metafeed) {
       cb(new Error('expected metafeed argument'))
     } else if (typeof metafeed === 'function') {
       cb(new Error('expected metafeed argument and visit argument'))
     } else {
-      const cb = maybeCB
       filter(metafeed, visit, (err, feeds) => {
         if (err) return cb(err)
         if (feeds.length === 0) return cb(null, null)
@@ -79,15 +81,24 @@ exports.init = function (sbot, config) {
       if (detailsErr) return cb(detailsErr)
 
       const { keys, seed } = metafeed
-      const { feedpurpose, feedformat, metadata } = details
+      const { feedpurpose, feedformat, metadata, recps, encryptionFormat } =
+        details
+      if (recps && (recps.length !== 1 || !isCloakedMsgId(recps[0]))) {
+        return cb(
+          new Error('metafeed encryption currently only supports groupId')
+        )
+      }
+
       const opts = sbot.metafeeds.messages.optsForAddDerived(
         keys,
         feedpurpose,
         seed,
         feedformat,
-        metadata
+        metadata,
+        recps,
+        encryptionFormat
       )
-      sbot.db.create(opts, (err) => {
+      sbot.db.create(opts, (err, msg) => {
         if (err) return cb(err)
         cb(null, sbot.metafeeds.query.hydrateFromCreateOpts(opts, seed))
       })
@@ -221,6 +232,7 @@ exports.init = function (sbot, config) {
       seed,
       keys,
       metadata: {},
+      recps: null,
     }
   }
 
--- a/messages.js
+++ b/messages.js
@@ -20,10 +20,17 @@ const { deriveFeedKeyFromSeed } = require('./keys')
  * Low level API for generating messages
  */
 exports.init = function init(sbot) {
-  function optsForAdd(mfKeys, feedKeys, nonce, feedpurpose, metadata) {
-    const type = nonce ? 'metafeed/add/derived' : 'metafeed/add/existing'
+  function optsForAdd(
+    mfKeys,
+    feedKeys,
+    nonce,
+    feedpurpose,
+    metadata,
+    recps,
+    encryptionFormat
+  ) {
     const content = {
-      type,
+      type: nonce ? 'metafeed/add/derived' : 'metafeed/add/existing',
       feedpurpose,
       subfeed: feedKeys.id,
       metafeed: mfKeys.id,
@@ -33,6 +40,7 @@ exports.init = function init(sbot) {
     }
 
     if (nonce) content.nonce = nonce
+    if (recps) content.recps = recps
 
     if (metadata) Object.assign(content, metadata)
 
@@ -41,7 +49,8 @@ exports.init = function init(sbot) {
       keys: mfKeys,
       contentKeys: feedKeys, // see ssb-bendy-butt/format.js
       content,
-      encryptionFormat: 'box2', // in case metadata.recps is set
+      recps,
+      encryptionFormat: encryptionFormat || 'box2',
     }
   }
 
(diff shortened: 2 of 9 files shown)


api.js
```diff
@@ -16,19 +17,20 @@ function detailsToVisit(details) {
...
-      deepEqual(feed.metadata, details.metadata || {})
+      (details.metadata ? deepEqual(feed.metadata, details.metadata) : true) &&
+      (details.recps ? deepEqual(feed.recps, details.recps) : true)
     )
   }
 }
 
 exports.init = function (sbot, config) {
   function filter(metafeed, visit, maybeCB) {
+    const cb = maybeCB
```
mixmix, line 28: I don't know why this code was written like this. I moved this.... either maybeCB isn't there and it errors, or... if I leave the code it definitely errors? Slight improvement

package.json
```diff
@@ -25,8 +25,7 @@
     "pull-defer": "^0.2.3",
     "pull-notify": "^0.1.1",
     "pull-stream": "^3.6.14",
-    "ssb-bfe": "^3.6.0",
-    "ssb-db2": ">=3.0.0 <=6",
+    "ssb-bfe": "ssbc/ssb-bfe#cloaked",
```
mixmix, line 28: NOTE - depends on this being merged

api.js
```diff
@@ -43,12 +45,12 @@ exports.init = function (sbot, config) {
   }
 
   function find(metafeed, visit, maybeCB) {
+    const cb = maybeCB
```
staltz, line 48: Yeah it seems the logic for cb here is all upside down. It's supposed to be like `create()` (line 70) because the arguments are flexible and we don't know if we have a callback. I think `find` and `filter` used to be public APIs but now that they're private, we can perhaps assume that the arguments are non-flexible (i.e. cb is always given, and always in the 3rd slot).

Note however that this is completely unrelated to your PR. Don't work on refactoring this in this PR.

query.js
```diff
@@ -61,16 +62,8 @@ exports.init = function (sbot, config) {
...
-        'subfeed',
-        'nonce',
-        'metafeed',
-        'tangles',
-        'type',
-      ]
-      for (const key of Object.keys(content)) {
-        if (ignored.includes(key)) continue
+      for (const key in content) {
+        if (NOT_METADATA.has(key)) continue
```
staltz, line 66: I think your PR #89 is going to grow more, so we could probably take these changes which are unrelated (they're left over from the reserved-keywords PR) and make it its own PR.

test/api/advanced/find-or-create.test.js
```diff
@@ -151,3 +151,81 @@ test('advanced.findOrCreate (protected metadata fields)', (t) => {
...
+
+          t.deepEqual(msgs[0].value.content.recps, [groupId], 'metafeed/add/derived has recps')
+          // t.deepEqual(typeof msgs[0].value?.meta?.originalContent, 'string', 'metafeed/add/derived is encrypted')
+
+          cb(null)
+        }))
+      }
+    ]
+
+    waterfall(series, cb)
```
staltz, line 197: Could we not add this dependency? I want the tests to run only "raw JS" code and *our* library, to rule out mistakes from other libraries and test only our library. If I understand correctly, we can use just a traditional callback pyramid for this purpose. It's not pretty, but everyone understands what's going on, unlike a new dependency.

staltz (changes requested):

test/api/advanced/find-or-create.test.js
mixmix: The pyramid is annoying to maintain IMO.
Agree we should discuss "standard tools" we're using. Will bring this up in next meeting and change this for now

api.js
mixmix: I had a test fail because of this, so I fixed it.
I would like some tooling that catches broken code like this (my linting/highlighting sees it in my editor)

mixmix: This is an example of something I'm not going to revert / split out.
It's not a good use of time and the code was broken as it was

query.js
mixmix: they are related. I had to add `recps` to this `ignored` list. So it was add recps as a string to this file, or swap this NOT_METADATA in.

I don't intend to change this one

test/api/advanced/find-or-create.test.js
mixmix: DONE

mixmix:
@staltz have made changes I'm going to, and highlighted the ones I'm not (along with reasons)
From here, can you review and merge.

You'll see I've pushed back on your requests to put some things in other PRs. I'd like you to consider this as an example of a place for you to concede some flexibility. I've done a lot of flexing my working approach to meet you, and it would feel good if we could both practice that.

mixmix:
Oh, NOTE I'm not working on supporting encryption to FeedId at all here. That's for a future PR (maybe, if we need it).
This change gives any user putting a FeedId in a clear error

api.js
staltz: I wasn't asking you to do a refactor, I was just suggesting about a new refactor, i.e.

> we can perhaps assume that the arguments are non-flexible

but I was advising to not do this new suggestion in this PR.

messages.js
```diff
@@ -20,10 +20,17 @@ const { deriveFeedKeyFromSeed } = require('./keys')
...
+    feedKeys,
+    nonce,
+    feedpurpose,
+    metadata,
+    recps,
+    encryptionFormat
+  ) {
     const content = {
-      type,
+      type: nonce ? 'metafeed/add/derived' : 'metafeed/add/existing',
```
staltz, line 33: @mixmix There's no action point in what I'm going to say, I just want to comment on a pattern that I find annoying. This refactor right here is merely stylistic. I wrote this part and I prefer to have the `const type`, and you just ripped that out, probably because it's not your style, without justifying it. Can I ask you to avoid doing these kind of changes that are (1) unjustified, (2) merely code style that conflicts with the original author's preference.

test/api/advanced/find-or-create.test.js
```diff
@@ -151,3 +151,76 @@ test('advanced.findOrCreate (protected metadata fields)', (t) => {
...
+        },
+        (err, f) => {
+          if (err) t.error(err, 'no error')
+
+          t.deepEqual(f.recps, [groupId], 'FeedDetails contains recps')
+
+          sbot.db.query(where(type('metafeed/add/derived')), toCallback((err, msgs) => {
+            if (err) return cb(err)
+
+            if (msgs.length !== 1) t.equal(msgs.length, 1, 'only one metafeed/add/derived')
```
staltz, line 186: I'm confused, isn't there supposed to be 2 `metafeed/add/derived` msgs? 1 for the root mf adding a shard, and 1 for the shard feed adding the group feed.

staltz: Also, why the if condition?

```suggestion
            t.equal(msgs.length, 1, 'only one metafeed/add/derived')
```

staltz (changes requested):

messages.js
mixmix: This is fair critique. I have a "tidy as I go" approach but we don't agree what "tidy" means.
I'll do my best to not do this (I did consciously leave other areas alone)

package.json
```diff
@@ -25,8 +25,7 @@
     "pull-defer": "^0.2.3",
     "pull-notify": "^0.1.1",
     "pull-stream": "^3.6.14",
-    "ssb-bfe": "^3.6.0",
-    "ssb-db2": ">=3.0.0 <=6",
+    "ssb-bfe": "^3.6.1",
```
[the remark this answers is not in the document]
mixmix: nice catch. I just did `npm i ssb-db2@latest` after publishing a new version.
Didn't realise this side effect

Committing revert

test/api/advanced/find-or-create.test.js
mixmix: Good question - I think this is a mistake in the tests to only expect 1. I wonder if it's passing because of a race condition around indexing (sigh).

I'm noticing this is quite common (if it's what it is) - to the point that I want a helper function. e.g.

```js
ssb.db.indexesReady(cb)
```

will investigate and report back

mixmix:
@staltz we agreed to work together towards a shared pattern - that I would push hard for micro-PRs (which I think I've been doing fairly well on), and that you would try flexing if there was something that was a minor step off that (I'm measuring by whether it changes the PR much logically, conceptually, thematically). 

Anyway that's for concession, it feels good for momentum.

Also thanks for the close review - you caught some good potential bugs
:heart:

test/api/advanced/find-or-create.test.js
staltz: For jitdb, everything you query should be up-to-date with the log. But for leveldb indexes, there is `ssb.db.onDrain(nameOfTheLevelIndex, cb)` as a way of waiting until it's indexed

mixmix:
oops I clicked the "Refresh" icon next to your name which was actually "request another review" (I must have some compulsion to click refresh)

test/api/advanced/find-or-create.test.js
mixmix: OOOOH, the reason there's only 1 `metafeed/add/derived` is because this is testing the lower level (advanced) API which doesn't do sharding

PR #90, opened 2022-10-05
branchStream outputs rootDetails (instead of null)

mixmix:
Problem: branchStream behaves differently than `findOrCreate` one returns details for root meta feed, the other doesn't.
This PR brings them into line


lookup.js
```diff
@@ -106,7 +115,10 @@ exports.init = function (sbot, config) {
 
     // Update roots
     if (!detailsLookup.has(metafeed)) {
-      detailsLookup.set(metafeed, null)
+      // !detailsLookup.has(metafeed) !== "metafeed is a root)
+      // this assumes that we have replicated the metafeed tree in the order
+      // that it was created.
```
mixmix, line 120: :fire: spotted a potential edge case that could cause us pain later.
Documented in code. Should this be an issue?

staltz: Yes, let's delete the commented code and create an issue.

staltz (changes requested):
Just a request to remove a code comment on the meta level. Then you can merge.

Other than that, I'm fine with merging this. I know why I originally chose `null` and I still would prefer the `null`: it's because there is no message that "created" this feed. The `details` refers to the contents of the message on a metafeed, and there is no message that declares a root metafeed.

Also, the object we're returning now is always the same object, so there's no information in it that we couldn't infer automatically. The root metafeed always has a bendybutt feedformat. I also find `feedpurpose === 'root'` misleading because it *seems* like we declared the root metafeed somewhere with `feedpurpose` set to `"root"` but that's not the case.

Overall, I'm on the whatever side with this PR, you can merge it if you like it.

mixmix:
I understand your perspective @staltz . I think as a consumer of the module I don't really think/ need to know about those details. All I was hitting was an annoying case where _sometimes_ I had to catch `null` and not dig into an object. That made my code slightly more annoying

PR #92, opened 2022-10-10
ensure that findOrCreate / findAndTombstone is never concurrent

staltz:
## Context

I'm working on updating ssb-replication-scheduler to use the new API in ssb-meta-feeds, and in the tests there are some cases of index feeds that are created at the same time.

## Problem

If two `findOrCreate` calls happen concurrently ("at the same time"), then both calls may create a "v1" feed under the root, which leads to a situation like

```mermaid
graph TB;
  root --> v1a[v1] & v1b[v1]
```

This can also happen with shard feeds and leaf feeds.

Also, this concurrency problem can happen with `findAndTombstone`: there should be only one `metafeed/tombstone` message for a given subfeed.

## Solution

Use `mutexify` in `findOrCreate` and `findAndTombstone` to enforce that only one call can happen at a time. I was working on a more complex solution before, which would *allow* concurrency if you're creating unrelated feeds like `root/v1/7/chess` and `root/v1/7/tictactoe`, but I realized that subfeed creation doesn't have to be efficient (we're not going to be creating feeds at a high throughput), so I went with the nuclear (and simpler) solution of allowing only one `findOrCreate` at a time, no matter what the arguments are.

### DIFF
--- a/api.js
+++ b/api.js
@@ -5,6 +5,7 @@
 const run = require('promisify-tuple')
 const deepEqual = require('fast-deep-equal')
 const { isCloakedMsgId } = require('ssb-ref')
+const mutexify = require('mutexify')
 const debug = require('debug')('ssb:meta-feeds')
 const pickShard = require('./pick-shard')
 const { BB1, v1Details, NOT_METADATA } = require('./constants')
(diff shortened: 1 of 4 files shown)


arj03 (approved):
Looks good, I like the better safe than sorry approach here. Makes debugging a lot easier if we should run into something.

staltz:
Yup, let's not create problems for ourselves to debug in the future. :sweat_smile:

PR #94, opened 2022-10-10
main located at root/v1/:shard/main branch

staltz:
## Context

I'm working on ssb-replication-scheduler to make it understand the v1 tree structure, and it expects all feeds to be in paths `root/v1/:shard/___`.

## Problem

ssb-meta-feeds was still doing `root/main`. See also https://github.com/ssbc/ssb-meta-feeds-spec/issues/35

## Solution

When the root metafeed is created, publish the main as a `metafeed/add/existing` under `root/v1/:shard/main`.

### DIFF
--- a/api.js
+++ b/api.js
@@ -132,7 +132,7 @@ exports.init = function (sbot, config) {
     findOrCreate(root, v1Visit, v1Details, cb)
   }
 
-  function findShard(v1Feed, root, details, cb) {
+  function findShard(root, v1Feed, details, cb) {
     const shardDetails = {
       feedpurpose: pickShard(root.keys.id, details.feedpurpose),
       feedformat: BB1,
@@ -141,7 +141,7 @@ exports.init = function (sbot, config) {
     find(v1Feed, shardVisit, cb)
   }
 
-  function findOrCreateShard(v1Feed, root, details, cb) {
+  function findOrCreateShard(root, v1Feed, details, cb) {
     const shardDetails = {
       feedpurpose: pickShard(root.keys.id, details.feedpurpose),
       feedformat: BB1,
@@ -225,7 +225,7 @@ exports.init = function (sbot, config) {
       const [err3, v1Feed] = await run(findOrCreateV1)(mf)
       if (err3) return release(cb, err3)
       const d = { feedpurpose: 'main' }
-      const [err4, shardFeed] = await run(findOrCreateShard)(v1Feed, mf, d)
+      const [err4, shardFeed] = await run(findOrCreateShard)(mf, v1Feed, d)
       if (err4) return release(cb, err4)
       const visit = (f) => f.feedpurpose === 'main'
       const [err5, added] = await run(find)(shardFeed, visit)
@@ -273,7 +273,7 @@ exports.init = function (sbot, config) {
       findOrCreateV1(rootFeed, (err, v1Feed) => {
         if (err) return cb(err)
 
-        findOrCreateShard(v1Feed, rootFeed, details, (err, shardFeed) => {
+        findOrCreateShard(rootFeed, v1Feed, details, (err, shardFeed) => {
           if (err) return cb(err)
 
           findOrCreate(shardFeed, detailsToVisit(details), details, cb)
@@ -291,7 +291,7 @@ exports.init = function (sbot, config) {
       findOrCreateV1(rootFeed, (err, v1Feed) => {
         if (err) return cb(err)
 
-        findShard(v1Feed, rootFeed, details, (err, shardFeed) => {
+        findShard(rootFeed, v1Feed, details, (err, shardFeed) => {
           if (err) return cb(err)
           if (!shardFeed) return cb(null, false)
 
--- a/api.js
+++ b/api.js
@@ -128,6 +128,28 @@ exports.init = function (sbot, config) {
     }
   }
 
+  function findOrCreateV1(root, cb) {
+    findOrCreate(root, v1Visit, v1Details, cb)
+  }
+
+  function findShard(v1Feed, root, details, cb) {
+    const shardDetails = {
+      feedpurpose: pickShard(root.keys.id, details.feedpurpose),
+      feedformat: BB1,
+    }
+    const shardVisit = detailsToVisit(shardDetails)
+    find(v1Feed, shardVisit, cb)
+  }
+
+  function findOrCreateShard(v1Feed, root, details, cb) {
+    const shardDetails = {
+      feedpurpose: pickShard(root.keys.id, details.feedpurpose),
+      feedformat: BB1,
+    }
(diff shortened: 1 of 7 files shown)


api.js
```diff
@@ -199,16 +221,22 @@ exports.init = function (sbot, config) {
...
+      const [err4, shardFeed] = await run(findOrCreateShard)(v1Feed, mf, d)
+      if (err4) return release(cb, err4)
+      const visit = (f) => f.feedpurpose === 'main'
+      const [err5, added] = await run(find)(shardFeed, visit)
+      if (err5) return release(cb, err5)
       if (!added) {
-        debug('adding main feed to root meta feed')
-        const opts = optsForAddExisting(mf.keys, 'main', config.keys)
+        debug('adding main feed to a shard metafeed')
+        const opts = optsForAddExisting(shardFeed.keys, 'main', config.keys)
```
mixmix, line 235: :fire: / :hot_pepper: 

If I understand this right, this is announcing the location of the feed structure on the classic `main` feed, right?
If that's the case I would like to recommend that we still announce the `rootMF`, as it's much more important to know than some shard feed right?

It might be that that's not really how `metafeed/add/existing` works, but we should perhaps consider include the rootMF as meta-data then...

api.js
```diff
@@ -128,6 +128,28 @@ exports.init = function (sbot, config) {
...
+  function findShard(v1Feed, root, details, cb) {
+    const shardDetails = {
+      feedpurpose: pickShard(root.keys.id, details.feedpurpose),
+      feedformat: BB1,
+    }
+    const shardVisit = detailsToVisit(shardDetails)
+    find(v1Feed, shardVisit, cb)
+  }
+
+  function findOrCreateShard(v1Feed, root, details, cb) {
```
mixmix, line 144: `<style>`

I would recommend shuffling the args to match the path _vibe_ of `root/v1/:shard/subfeedDetails`, i.e.

```js
function findOrCreateShard(root, v1Feed ,details, cb) {
```

package.json
```diff
@@ -49,7 +49,7 @@
     "test": "npm run test:js && npm run test:only",
     "test:js": "tape \"test/**/*.test.js\" | tap-arc --bail",
     "test:only": "if grep -r --exclude-dir=node_modules --exclude-dir=coverage --exclude-dir=.git --color 'test\\.only' .; then exit 1; fi",
-    "format-code": "prettier --write \"*.js\" \"test/*.js\"",
+    "format-code": "prettier --write \"*.js\" \"test/**/*.js\"",
```
mixmix, line 52: ## :ghost: 

NICE CATCH

test/api/branch-stream.test.js
```diff
@@ -16,32 +16,62 @@ test('branchStream', (t) => {
...
-        t.equal(branches.length, 5, '5 branches')
-
-        t.equal(branches[0].length, 1, 'root mf alone')
-        t.equal(typeof branches[0][0][0], 'string', 'root mf alone')
+        t.equal(branches.length, 7, '7 branches')
+        const [
+          root,
+          rootV1,
+          rootV1Shard,
+          rootV1ShardMain,
```
mixmix, line 24: I looked at this thinking it was wrong for a long time.
I think a comment like "note rootChess etc were all added using the advanced API, skipping the v1/shard pattern"

test/api/branch-stream.test.js
```diff
@@ -92,21 +128,32 @@ test('branchStream (encrypted announces)', (t) => {
...
+        pull.collect((err, branches) => {
+          if (err) t.error(err, 'no error')
+
+          t.equal(branches.length, 6, '6 feed branches')
+          // root
+          // root/v1
+          // root/v1/:shardA
+          // root/v1/:shardA/main
+          // root/v1/:shardB
+          // root/v1/:shardB/dental
```
mixmix, line 143: :heart:
these help a lot

test/api/find-or-create.test.js
```diff
@@ -81,7 +121,12 @@ test('double findOrCreate should not create two v1 feeds', async (t) => {
   const shardAnnouncements = msgs.filter((msg) =>
     hexes.includes(msg.value.content.feedpurpose)
   )
-  t.equals(shardAnnouncements.length, 1, 'only one shard announcement')
+  // sometimes :shardA and :shardB are the same, among
+  // root/v1/:shardA/main and root/v1/:shardB/chess
```
mixmix, line 125: :hankey: 

Ideally we have a way to deterministically create the root mf keys so we get a good test for things like this.
I'm happy know this would intermitently fail if there was something genuinely wrong, so this seems like a fine solution

mixmix (changes requested):
Only one thing I think we "must" consider before merging (see :fire: / :hot_pepper:)

Feedback: there were two changes made in this PR which could have been seperated into 2 PRs to smooth reviewing:
- move the main subtree to `root/v1/:shard/main`
- prettier changes (this changed the formatting of lots of tests which forced me to read closely because you were also changing the API at the same time)

_**Tone**: I'm not being smug, I'm genuinely trying to embody what I think we said we both wanted to value (PRs that are easy to review and merge, and code that doesn't have regressions). I am enjoying noticing the pain you've pointed out to me. Interested if you're giving me a taste of it to inoculate me, or if you didn't notice yourself :rofl: :heart:_

This is a good PR, love this tuning
@arj03 we need to make sure this change is recorded in the "migration" spec

api.js
staltz: I think you understood it the other way around. What this line does is that it announces the main feed as a subfeed of the shard, by publishing a message on the shard.

api.js
staltz: Yeah that's a nice one.

test/api/find-or-create.test.js
staltz: Yeah, I also feel poopy about this. We should definitely add a seed for the root, for tests, but it's a bit more elaborate than this PR can fit. Should/could be done soon.

staltz:
> Feedback: there were two changes made in this PR which could have been seperated into 2 PRs to smooth reviewing:

Yeah, you're right

> we need to make sure this change is recorded in the "migration" spec

Indeed!

test/api/branch-stream.test.js
staltz: Done

api.js
mixmix: Oh I thought that would just be a meta/add/existing on the sharf feed which points to a  main feed.

I'm assuming there is a message published on the main feed that points back to the root feed (to support people migrating)

staltz: Yes there is, check the code block above this block.

mixmix:
If I misunderstood that announce thing that's fine. Main thing I wanted to check was that we hadn't missed a side effect of changing where and how we attach the main feed.

The mobile ui sucks, so:

# approve :+1:

PR #95, opened 2022-10-10
tidy the API results

staltz:
## Problem

There are a couple legacy decisions in this library that are still haunting us. Like `feedpurpose` instead of `purpose`, it's cemented in the spec and not worth changing it. But also `details.subfeed`. The worst is when `details` refers to the root metafeed, and thus `details.subfeed` is the **id of the root metafeed**. Disgusting! Someone will definitely feel confused by this API.

Also, the branches as tuples `[id, details]` was getting really obnoxious to work with, when it could just be `details.id` etc.

## Solution

We can keep the spec untouched while making this library return more reasonable.

- `details.metafeed` => `details.parent`
- `details.subfeed` => `details.id`
- `details.feedpurpose` => `details.purpose`
- `details.feedformat` => `details.feedFormat` (this one can't be `format` because we also have `details.encryptionFormat`)

:bomb: This is a breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ We create a subfeed for `about` messages under our `root` feed using
 subfeed that matches the criteria.
 
 ```js
-const details = { feedpurpose: 'aboutMe' }
+const details = { purpose: 'aboutMe' }
 sbot.metafeeds.findOrCreate(details, (err, aboutMeFeed) => {
   console.log(aboutMeFeed)
 
@@ -97,7 +97,7 @@ Once you have a *FeedDetails* object, like `aboutMeFeed`, you can publish on
 the new subfeed:
 
 ```js
-const details = { feedpurpose: 'aboutMe' }
+const details = { purpose: 'aboutMe' }
 sbot.metafeeds.findOrCreate(details, (err, aboutMeFeed) => {
   console.log(aboutMeFeed)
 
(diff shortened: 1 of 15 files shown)


mixmix (approved):
This is music to my ears.
Great idea

Have reviewed code on phone so not the closest read but trust any problems surfaced are easy to solve and trust our tests

README.md
```diff
@@ -136,10 +136,10 @@ Arguments:
 - `cb` *function* delivers the response, has signature `(err, FeedDetails)`, where FeedDetails is
     ```js
     {
-      metafeed: 'ssb:feed/bendybutt-v1/sxK3OnHxdo7yGZ-28HrgpVq8nRBFaOCEGjRE4nB7CO8=',
-      subfeed: '@I5TBH6BuCvMkSAWJXKwa2FEd8y/fUafkQ1z19PyXzbE=.ed25519',
-      feedpurpose: 'chess',
-      feedformat: 'classic',
+      id: '@I5TBH6BuCvMkSAWJXKwa2FEd8y/fUafkQ1z19PyXzbE=.ed25519',
```
mixmix, line 139: at some point we should consider just using `ssb-uri` as default way to communicate feedIds

lookup.js
```diff
@@ -106,38 +107,39 @@ exports.init = function (sbot, config) {
       details.tombstoned = true
       details.reason = content.reason
     }
-    details.recps = content.recps || null
     return details
   }
 
   function updateLookup(msg) {
     const { type, subfeed, metafeed } = msg.value.content
```
mixmix, line 114: style:

```js
const { type, subfeed: id, metafeed: parent } = msg.value.content
```

mixmix (approved):
LOVE IT.

couple of style/opinion suggestions, but meh.

I think if we're going for "micro PRs" this looks like it could have been split into 2:
- change details pattern
- change the branchStream output

It would have made it easier to review, but I'm also fine with this

lookup.js
staltz: I secretly hate this kind of destructuring because I'm never sure if it's the left side being renamed or the right side, but :shrug: okay

PR #96, opened 2022-10-13
ensure not box1 dependant

mixmix:
Depends on : https://github.com/ssbc/ssb-db2/pull/397

### DIFF
--- a/messages.js
+++ b/messages.js
@@ -193,13 +193,13 @@ exports.init = function init(sbot) {
      */
     optsForSeed(mfKeys, mainfeedId, seed) {
       return {
-        recps: [mainfeedId],
-        encryptionFormat: 'box',
         content: {
           type: 'metafeed/seed',
           metafeed: mfKeys.id,
           seed: seed.toString('hex'),
         },
+        recps: [mainfeedId],
+        encryptionFormat: 'box2',
       }
     },
   }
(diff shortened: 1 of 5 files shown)


messages.js
```diff
@@ -193,13 +193,13 @@ exports.init = function init(sbot) {
...
       return {
-        recps: [mainfeedId],
-        encryptionFormat: 'box',
         content: {
           type: 'metafeed/seed',
           metafeed: mfKeys.id,
           seed: seed.toString('hex'),
         },
+        recps: [mainfeedId],
+        encryptionFormat: 'box2',
```
mixmix, line 202: :question:  do we want to lock ourselves to one encryptionFormat like this?

note that I don't want it to be `box` by default because I don't plan to use that in my apps.

test/testbot.js
```diff
@@ -22,8 +22,10 @@ module.exports = function createSbot(opts = {}) {
   const keys = opts.keys || ssbKeys.loadOrCreateSync(path.join(dir, 'secret'))
 
   const stack = SecretStack({ appKey: caps.shs })
-    .use(require('ssb-db2'))
+    .use(require('ssb-db2/core'))
```
mixmix, line 25: the README says that you can use `ssb-db2` >= 5, the dependences say > 3...
do all of these have the latest `ssb-box2` in them by default?

I know they don't because I'm about to open a PR updating the 6...

messages.js
staltz: @arj03 Do you have opinions about encrypting the root mf seed as box1 or box2? I don't have an opinion, but I know you're the one who devised this solution.

test/testbot.js
staltz: Hmm, yeah feel free to update these. I don't think there's a point in supporting older ssb-db2 versions, because ssb-meta-feeds is so new.

staltz (approved):
:monkey:

PR #97, opened 2022-10-13
flexy definition of group recp

mixmix:
hit a problem, needed this

### DIFF
--- a/api.js
+++ b/api.js
@@ -86,6 +86,10 @@ exports.init = function (sbot, config) {
 
       const { keys, seed } = metafeed
       const { purpose, feedFormat, metadata, recps, encryptionFormat } = details
+      // NOTE - we check the sorts of encryption we're allowing because we don't
+      // currently have a clear spec for:
+      //   - what "self encryption" means in meta-feeds
+      //   - when you encrypt to someone else, *which feedId* do you use?
       if (recps && (recps.length !== 1 || !isGroupId(recps[0]))) {
         return cb(
           new Error('metafeed encryption currently only supports groupId')
--- a/api.js
+++ b/api.js
@@ -6,9 +6,12 @@ const run = require('promisify-tuple')
 const deepEqual = require('fast-deep-equal')
 const { isCloakedMsgId } = require('ssb-ref')
 const mutexify = require('mutexify')
-const debug = require('debug')('ssb:meta-feeds')
+const { keySchemes } = require('private-group-spec')
+
 const pickShard = require('./pick-shard')
 const { BB1, v1Details, NOT_METADATA } = require('./constants')
+
+const debug = require('debug')('ssb:meta-feeds')
 
 const alwaysTrue = () => true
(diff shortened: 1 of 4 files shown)


staltz (changes requested):
What if we drop this recps validation (on line 87) entirely? It doesn't seem like it belongs to this module. `recps` and `encryptionFormat` are just passed along down to ssb-db2 which then passes `recps` to ssb-box2, and it's ssb-box2's responsibility to validate recps.

It feels out of place that we're doing box2 recps validation in this module which should just be about the metafeed tree. Perhaps we could add this validation to ssb-tribes2 once that one uses ssb-meta-feeds. 

It would be useful to allow box1 encryption here too. Use case: I create a leaf feed for Ferment (by definition an old app) and suppose I want to do box1 private messages there. ssb-meta-feeds shouldn't be stopping me from doing that.

mixmix:
@staltz this is a good point. I should have left a note in the code (and now have).
The reason we're checking recps is because we don't have a clear specification for how non-groupId recps should work. i.e. if I put a `feedId` in... tell me how I use that to map to a `keyScheme` in meta-feeds? Should I always use the rootMF? If we don't then we will have to try 100x more decryptions because any encrypted message could be DM'ing ANY of our subfeeds.

UNDER-specified => locked down 

Recap from https://github.com/ssbc/ssb-meta-feeds/pull/89
image

mixmix:
Opened an issue https://github.com/ssbc/ssb-meta-feeds-migration/issues/4

staltz:
Mmm, sorry Mix, I still don't understand why this logic would belong in this module.

> if I put a feedId in... tell me how I use that to map to a keyScheme in meta-feeds? Should I always use the rootMF?

Suppose you're using ssb-meta-feeds inside a module called ssb-foo-bar. Then the recps selected inside ssb-foo-bar are passed to ssb-meta-feeds, which just redirects them down to ssb-box2, which in turn *does some validation* and then redirects them to ssb-keyring, which also does *some validation* albeit different kind of validation (e.g. can't do a DM between two keys you own).

```mermaid
graph LR;
  ssb-foo-bar --> ssb-meta-feeds --> ssb-box2 --> ssb-keyring
```

My point is that validation should happen in `ssb-box2` and `ssb-keyring`, and that it's `ssb-foo-bar`'s responsibility for choosing the friend's rootMF if they want to *announce* a private subfeed that only the friend's rootMF can know about. So I'm saying that ssb-meta-feeds shouldn't do any recps validation.

> UNDER-specified => locked down

What does this mean?

staltz:
For the sake of getting things done, I pushed commit https://github.com/ssbc/ssb-meta-feeds/commit/d894b1ed6a0911e06e48dd42ecfd3c24f184edac to master which removes validation of recps. This PR was last commented on 16 days ago, and we have to move a lot faster than that, so I just made a decision.

We can keep discussing this idea and it's possible to revert the decision in the future if we want.

PR #98, opened 2022-10-20
add seed config to make tests predictable

staltz:


api.js
```diff
@@ -25,6 +25,9 @@ function detailsToVisit(details) {
 }
 
 exports.init = function (sbot, config) {
+  if (!config) config = {}
```
mixmix, line 28: `<style>` - any particular reason you don't like?

```
exports.init = function (sbot, config = {}) {
```
Just want to know in case there is a hole I might not know about that I'm gonna fall in because I do it this way all the time

api.js
```diff
@@ -25,6 +25,9 @@ function detailsToVisit(details) {
...
+  if (!config.metafeeds) config.metafeeds = {}
```
mixmix, line 29: :fire: / `style`
I would say it's a very bad idea to have plugins mutating the `config` - it's probably a bigger issue with SecretStack that there's no scoped config. hmmm

package.json
```diff
@@ -43,7 +43,7 @@
     "ssb-box2": "^3.0.0",
     "ssb-caps": "^1.1.0",
     "ssb-classic": "^1.1.0",
-    "ssb-db2": "ssbc/ssb-db2#encryptionFormat_grrr",
+    "ssb-db2": "6.2.4",
```
mixmix, line 46: haha _woops_

test/testbot.js
```diff
@@ -31,5 +31,11 @@ module.exports = function createSbot(opts = {}) {
   return stack({
     path: dir,
     keys,
+    metafeeds: {
+      seed: Buffer.from(
+        '000000000000000000000000000000000000000000000000000000000000beef',
```
mixmix, line 36: I love little easter-eggs like this

# :rofl: :heart:

mixmix (approved):
LOVE IT

There is a :fire: / style in here which is a dangerous pattern IMO,  but the actual mutation you performed seems safe enough to merge.

api.js
mixmix: OH - we could just deepFreeze the config Object after it's passed in.
I would feel real happy about that.

api.js
staltz: No particular reason, JS often has different ways of achieving stuff

PR #100, opened 2022-10-21
fix findAndTombstone to not use mutation

staltz:
## Context

My tests in ssb-replication-scheduler were failing on findAndTombstone

## Problem

There are two problems:

1. `feedformat` is supposed to be `feedFormat`
2. findAndTombstone mutates the input `details` to add `.feedformat` to it, so if you pass the same `details` object (like our tests do), it'll work, but if you pass two different objects that are deep equal, it won't work

## Solution

Rename `feedformat` to `feedFormat` and don't mutate the inputs.

### DIFF
--- a/api.js
+++ b/api.js
@@ -266,7 +266,7 @@ exports.init = function (sbot, config = {}) {
       return
     }
 
-    if (!details.feedFormat) details.feedFormat = 'classic'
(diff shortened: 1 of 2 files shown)


mixmix (approved):

mixmix:
good catch

PR #103, opened 2022-11-07
branchStream can be called immediately after findOrCreate

staltz:
## Context

Tests were failing downstream in ssb-tribes2 when I created a groupFeed and then tried to use branchStream to replicate that to other peers.

## Problem

After findOrCreate, branchStream had to "wait" for the ssb-db2 bendybutt message to be processed *and then* converted to a "details" object.

If you called branchStream immediately after findOrCreate was done, this means you would miss the newly-created leaf feed.

## Solution

As soon as findOrCreate is done, `updateLookup` on the branchStream logic so that the data is synchronously available there.

### DIFF
--- a/lookup.js
+++ b/lookup.js
@@ -91,15 +91,15 @@ exports.init = function (sbot, config) {
     }
   }
 
-  function msgToDetails(prevDetails, msg) {
+  function msgToDetails(msg) {
     const content = msg.value.content
-    const details = { ...prevDetails }
-    details.id = prevDetails ? prevDetails.id : content.subfeed
-    details.parent = content.metafeed || details.parent
-    details.purpose = content.feedpurpose || details.purpose
+    const details = {}
+    if (content.subfeed) details.id = content.subfeed
+    if (content.metafeed) details.parent = content.metafeed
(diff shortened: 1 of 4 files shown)


lookup.js
```diff
@@ -91,15 +92,15 @@ exports.init = function (sbot, config) {
...
-    details.parent = content.metafeed || details.parent
-    details.purpose = content.feedpurpose || details.purpose
+    const details = {}
+    if (content.subfeed) details.id = content.subfeed
+    if (content.metafeed) details.parent = content.metafeed
+    if (content.feedpurpose) details.purpose = content.feedpurpose
     details.feedFormat = validate.detectFeedFormat(content.subfeed)
     details.recps = content.recps || null
-    details.metadata = {} || details.metafeed
+    details.metadata = {} || details.metadata
```
arj03, line 103: here details.metadata will always be undefined because you start details empty

staltz: Oh shoot

arj03:
Just one minor thing, the rest looks good and great way to simplify the usage of this module.

arj03 (approved):
Great :+1:

PR #104, opened 2022-11-09
add .keys to all feed details you own

staltz:
## Context

In ssb-box2 we're going to have to traverse "my" metafeed tree, and then match that with other peers' metafeed tree to create DH keys.

## Problem

When using default branchStream, it returns any branch in the "forest" (the collection of all trees), and there isn't any way how we can know whether a branch belongs to _my_ tree, or some _other_ peer's tree.

## Solution

Add `.keys` field for all feed details returned by branchStream if I own that feed. If `details.keys` exists, it means you own the subfeed (well, by definition, because if you have the `keys`, you can publish to that feed). 

In the implementation, this requires loading our `seed` before querying ssb-db2 for messages that build up the branches lookup data structure. We also load the root metafeed keys in that data structure ASAP, so that whenever we stumble upon a feed, we can check if it's parents `details.keys` exists, and then hydrate the keys with the seed.

Asking for a review from @arj03, but pinging @mixmix just to be aware of this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,8 +196,9 @@ branch looks like this:
 ```
 
 Or in general, an `Array<Details>`. The *Details* object has
-the shape `{ id, purpose, feedFormat, metafeed, metadata }` like what
-`findOrCreate` returns.
+the shape `{ id, purpose, feedFormat, keys, parent, metadata }` like what
+`findOrCreate` returns. If the details is for a feed that doesn't belong to you,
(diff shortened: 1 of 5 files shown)


staltz:
Can haz eyeballs on this?

arj03 (approved):
Somehow missed this. Thanks for the ping. The changes looks good. Relatively easy to review.

PR #105, opened 2022-11-09
introduce FeedDetails class

staltz:
## Context

This is not a necessary addition, it's a nice-to-have addition.

I was thinking how we return `{ id, parent, purpose, ... }` objects in a bunch of places, and it's getting kind of out of hand. Several different files are creating/updating these objects, and they *have* to follow the same shape, but we're not enforcing the same shape.

Moreover, I'm also thinking that there will be many of these objects, because each branch in `branchStream` has many of these. It will be better to profile memory usage if we use classes, because they show up immediately in the Chrome DevTools. Also, it helps the V8 optimizer if all of the objects share exactly the same shape (same set of `Object.keys()` always) because then under the hood it creates a hidden `struct` (in C) to back all these objects.

You could say that's premature optimization, but the primary motivation here is just organization and code readability. It's a bonus that it just happens to be better for performance and memory profiling.

## Solution

Introduce a new class `FeedDetails` that can be constructed using various helpers: `fromMyMsg`, `fromCreateOpts`, etc. Replace all the hand-written feed details objects with this class instance.

There's a tiny breaking change with tombstoned feed details: `reason` => `tombstoneReason`. I can revert that if y'all have a different opinion.

Tests pass locally, but they don't run in CI yet because this PR isn't targeting `master`. This PR should be merged only after #104 is merged.

### DIFF
--- a/FeedDetails.js
+++ b/FeedDetails.js
@@ -1,0 +1,204 @@
+// SPDX-FileCopyrightText: 2022 Andre 'Staltz' Medeiros <contact@staltz.com>
+//
+// SPDX-License-Identifier: LGPL-3.0-only
+
+const { NOT_METADATA, BB1 } = require('./constants')
+const validate = require('./validate')
+const metafeedKeys = require('./keys')
+
+const correctPassword = Symbol('FeedDetails')
+
+function collectMetadata(content) {
+  const metadata = {}
+  for (const key in content) {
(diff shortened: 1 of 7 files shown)


arj03 (approved):
Great. I read through this and it makes sense. I like how this moves all the message handling into 1 file. About the tombstoneReason then fine by me. Better do it now than later.

PR #106, opened 2022-11-11
fix findAndTombstone tests

staltz:
The tests were using an outdated branchStream API shape and just happened to pass.


arj03 (approved):
👍

PR #107, opened 2022-11-11
Fix FeedDetails.equals to compare buffers correctly

staltz:
## Context

I got some weird duplicate branches coming out of `branchStream`, and it turned out that the '.equals()' function wasn't working here:

https://github.com/ssbc/ssb-meta-feeds/blob/895781aa18d789d79600020119bb22150061ef3a/lookup.js#L134

## Problem

`FeedDetails`'s `equals()` is too simplistic, it's comparing two buffers with `===`, and it's comparing two objects (`keys`) with `===` too.

## Solution

Compare buffers with `Buffer.equals` and compare objects with fast-deep-equal.

### DIFF
--- a/FeedDetails.js
+++ b/FeedDetails.js
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: LGPL-3.0-only
 
+const deepEqual = require('fast-deep-equal')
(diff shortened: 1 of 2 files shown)


arj03 (approved):
Ah yes ;) 👍

PR #110, opened 2022-11-15
fix branchStream to restart after box2 decryption

staltz:
## Context

I'm working on putting 2b funky recps in ssb-tribes2, and noticed a race condition in the tests.

## Problem

There's a race condition between:

1. Replicate the shard feed that announces (encrypted!) a group feed
2. Replicate the `group/add-member` message (on the invitations feed) that allows us to decrypt (1)

Ideally, (2) happens first and then (1), but if (1) happens first we have a problem. Even though ssb-tribes2 will run reindexEncrypted() when it bumps into a `group/add-member` msg, the `branchStream()` in ssb-meta-feeds won't know that it has to "restart". It just stays stuck in live mode, and of course there are no "new" messages. We only have old messages that were decrypted.

## Solution

This is not an ideal solution, but it works. I have an idea for the ideal case, which would be a new API in ssb-db2 that streams all the messages that were freshly decrypted. We need that not only for this branchStream problem, but in other cases too, like in the ssb-box2 "monitor" that matches DM triangles.

The current solution is to just track `indexingActive` (because reindexEncrypted() bumps that) and restart the branchStream query when indexingActive goes to zero. The live stream was also converted to old-and-live because that's the only way how we can get the old-but-freshly-decrypted messages, AND still get the live messages.

### DIFF
--- a/test/api/branch-stream.test.js
+++ b/test/api/branch-stream.test.js
@@ -4,8 +4,11 @@
 
 const test = require('tape')
 const pull = require('pull-stream')
+const ssbKeys = require('ssb-keys')
+const p = require('util').promisify
 const Testbot = require('../testbot')
 const { setupTree, testReadAndPersisted } = require('../testtools')
+const keysAPI = require('../../keys')
 
 test('branchStream', (t) => {
   const sbot = Testbot()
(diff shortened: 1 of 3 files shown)


test/api/branch-stream.test.js
```diff
@@ -138,3 +141,95 @@ test('branchStream (encrypted announces)', (t) => {
...
+    [metafeedKeys.id],
+    'box2'
+  )
+  await p(sbot.db.create)(opts)
+  t.pass('published encrypted announce message on a shard')
+
+  sbot.box2.setOwnDMKey(ownKey2)
+  t.pass('change own DM key so that branchStream cannot decrypt')
+
+  await new Promise((resolve) => {
```
arj03, line 195: Could you add one that is running live here and let that resolve once we have the one we are looking for. Just to check that live also works?

arj03:
Had a minor comment. Would be better with a solution in db2 so other modules doesn't need to be aware of this as you say. I think this solution is fine for now, lets get this in and improve this along the way. Just create an issue for it in db2 :)

test/api/branch-stream.test.js
staltz: Yes, I can. But i realized that adding the live mysteriously breaks the old and that doesn't sound right... Hmmm

staltz: https://github.com/ssbc/ssb-meta-feeds/pull/111

arj03 (approved):
Pre-approving this, the changes looks good after the other PR is merged

PR #111, opened 2022-11-17
fix branchStream "loadState" logic

staltz:
## Context

https://github.com/ssbc/ssb-meta-feeds/pull/110#discussion_r1023762878

## Problem

If you call branchStream with live first, and then immediately call branchStream with old second, then the second one will just assume that the state has loaded, without actually waiting for it.

## Solution

Just a bit of booleans and some ifs.


arj03 (approved):
Yep, makes sense

PR #112, opened 2022-12-20
Use db2 reindexed

arj03:
This is tidying up #110 with the new db2 reindexed api


lookup.js
```diff
@@ -171,30 +172,31 @@ exports.init = function (sbot, config) {
...
-                ),
-                pull.filter((msg) => validate.isValid(msg)),
-                (liveDrainer = pull.drain(updateLookupFromMsg))
-              )
-            })
+            pull(
+              sbot.db.reindexed(),
+              pull.filter((msg) => {
+                return (
+                  SSBURI.isBendyButtV1FeedSSBURI(msg.value.author) &&
```
staltz, line 184: We could either remove this check (because validate.isValid checks for msg.value.contentSignature which is duck-typing for bendybutt msgs) OR we could move this check into validate.isValid() itself

staltz (changes requested):
Sweet! Thank you! I have just one suggestion.

lookup.js
arj03: Okay, deleted the check.

Issue #117, opened 2023-02-12
Ordering of messages


arj03:
I'm looking at a trace that looks something like this:

```
notify branch [ 'root', 'v1' ] ssb:feed/bendybutt-v1/PVC3N4bEm5cVk5PMwZZ2BYI414B4kaqgxYyWsx29w5g=
got branch [ 'root', 'v1' ]
notify branch [ 'root', 'v1', '4' ] ssb:feed/bendybutt-v1/hqBQcbGAAKc5sG5oaJ4yzLwKipXmaVC6OmmOjFB---c=
got branch [ 'root', 'v1', '4' ]
notify branch [ 'root', 'v1', '4', 'main' ] @3XVgQP3VFF4r+YMtJk3QgOSz5zAWvfZXS0zYfqppf14=.ed25519
got branch [ 'root', 'v1', '4', 'main' ]
notify branch [ 'root', 'group/additions' ] @oH4x34UXqEkdUaJ2aY83+l6ZftNzJ6k7BU2hS8Edxkk=.ed25519
notify branch [ '1' ] ssb:feed/bendybutt-v1/TH9SxgrS2vHbp6iRsSlS3SaqFrIn-8-DzB4t-dNig5g=
got branch [ '1', 'group/additions' ]
got branch [ '1' ]
```

Got branch is what you get from `ssb.metafeeds.branchStream({ old: true, live: true })` and notify is from inside this module on live data. 

The tree looks like this:

```
root
└─┬ v1
  ├─┬ 4
  │ └── main
  └─┬ 1
    └── group/additions
```

For some reason it seems like 1 doesn't have a proper parent (v1), which is really strange because it seems like 4 works fine.

arj03:
And the second bug seems to be that the messages comes in the wrong order. `updateLookupFromMsg` first gets the group/additions and later the 1. So the parent links don't correctly align. It's a bit strange, because the messages from network should come in, in the correct order. So it's a bit strange that we would get them shuffled.

arj03:
Argh fuck. It's because `replicate` is cheating and just requesting all feeds at once. https://github.com/ssbc/ssb-tribes2/blob/4beef23cc85cd2fd4f34a1ab8e3f63bc3d54daba/test/helpers/replicate.js#L13

PR #118, opened 2023-02-13
Make sure FeedDetails.update also updates parent

arj03:
Not sure if the real problem is that messages come in out of order, but at least wanted to flag that this function should update parent in any case to be safe.

See this one for details: https://github.com/ssbc/ssb-meta-feeds/issues/117

### DIFF
--- a/FeedDetails.js
+++ b/FeedDetails.js
@@ -174,7 +174,7 @@ class FeedDetails {
    * Incorporates the given FeedDetails into this one, updating our properties.
    */
   update(feedDetails) {
-    this.parent = feedDetails.parent || this.parent
+    this.parent = this.parent || feedDetails.parent
     this.purpose = feedDetails.purpose || this.purpose
     this.metadata = { ...this.metadata, ...feedDetails.metadata }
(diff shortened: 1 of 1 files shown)


arj03:
It seems to work without this change (see https://github.com/ssbc/ssb-tribes2/pull/60). That is probably because now the order is correct now. Should we close this @staltz or what do you think?

FeedDetails.js
```diff
@@ -174,6 +174,7 @@ class FeedDetails {
    * Incorporates the given FeedDetails into this one, updating our properties.
    */
   update(feedDetails) {
+    this.parent = feedDetails.parent || this.parent
```
staltz, line 177: Instead of closing this PR, we could change this line to

```suggestion
    this.parent = this.parent || feedDetails.parent
```

Then I'd be much more comfortable. Because we don't want a subfeed to mutate its parent, that would be weird. But if it's null/undefined, then we do want to set it to a valid parent.

arj03: Right, thanks. Will merge after the tests pass

PR #119, opened 2023-02-15
Add findRootFeedId()

Powersource:
For https://github.com/ssbc/ssb-tribes2/pull/58/files#r1103913950


api.js
```diff
@@ -175,6 +175,15 @@ exports.init = function (sbot, config = {}) {
     })
   }
 
+  function findRootFeedId(subfeedId, cb) {
+    findById(subfeedId, (err, subFeed) => {
+      if (err) return cb(err)
+      // if err and subFeed are null then we've likely found root (can't findById root, at least atm)
+      if (subFeed === null) return cb(null, subfeedId)
```
mixmix, line 182: this feels dangerous IMO.
worst case, for some reason you have a subfeed and not it's parent feed, and you are like "oh can't find a parent, guess this is the root!"

I think the root feed should probably explicitly have some tag on it which is like "I AM ROOT"
OH, another easier and more safe option is that the last subfeedLink before you hit root has `purpose: "v1"`. It's not water-tight, but it is strong.

example:
```js
function findRootFeed (subFeed, cb) {
  findParentFeed(subFeed, (err, parentFeed) {
    if (err) return cb(err)

    if (parentFeed === null) return cb(null, subFeed) // done
    else findRootFeed(parentFeed, cb) // keep going
  }
}

function findParentFeed (subFeed, cb) {
  findById(subfeed.parent, (err, parentFeed) => {
    if (err) return cb(err)

    if (parentFeed) return cb(null, parentFeed)
    else {
      if (subFeed.purpose === "v1") return cb(null, null) // there is no parent, subFeed was root
      else return cb(new Error('Could not find parent feed')) 
    }
  })
}
```

The problem is that none of the code handles the case where you give it the root feed well. You would need to write another query like the one in `looup.js#findById` which does a query looking for the parent/metafeed fields...... should we do that?

yeah I'll try a PR which does that k

mixmix: HERE: https://github.com/ssbc/ssb-meta-feeds/pull/120
I've asked staltz to review as it's got some db2 query and I'm making a new index... which I think is fine, but want to check

mixmix:
@Powersource this one is good to go now

PR #120, opened 2023-02-16
add isRootFeedId, use in findRootFeedById

mixmix:
merges into https://github.com/ssbc/ssb-meta-feeds/pull/119

### DIFF
--- a/lookup.js
+++ b/lookup.js
@@ -21,16 +21,17 @@ const SSBURI = require('ssb-uri2')
 const validate = require('./validate')
 const FeedDetails = require('./FeedDetails')
 
-const SUBFEED_PREFIX_OFFSET = Math.max(
+const SUBFEED_PREFIX_OFFSET = (PARENTFEED_PREFIX_OFFSET = Math.max(
   '@'.length,
   'ssb:feed/bendybutt-v1/'.length,
   'ssb:feed/indexed-v1/'.length,
   'ssb:feed/gabbygrove-v1/'.length
-)
+))
 
 const B_VALUE = Buffer.from('value')
 const B_CONTENT = Buffer.from('content')
 const B_SUBFEED = Buffer.from('subfeed')
+const B_PARENT = Buffer.from('metafeed') // NOTE parent = metafeed in message content
 
 function seekSubfeed(buffer) {
   let p = 0 // note you pass in p!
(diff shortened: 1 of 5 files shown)


Powersource (commented):
cool

lookup.js
```diff
@@ -21,16 +21,17 @@ const SSBURI = require('ssb-uri2')
...
   'ssb:feed/bendybutt-v1/'.length,
   'ssb:feed/indexed-v1/'.length,
   'ssb:feed/gabbygrove-v1/'.length
-)
+))
 
 const B_VALUE = Buffer.from('value')
 const B_CONTENT = Buffer.from('content')
 const B_SUBFEED = Buffer.from('subfeed')
+const B_PARENT = Buffer.from('metafeed') // NOTE parent = metafeed in message content
```
staltz, line 34: Let's name this `B_METAFEED`, because these `B_` buffers are a convention that arj and I have been using in db2 and other modules where we know that `B_FOO` is just a shortcut for `Buffer.from('foo')`. If you create an exception for this rule, then it's suddenly not anymore guaranteed that `B_FOO` stands for `Buffer.from('foo')` and can create confusion i.e. debugging nightmares.

staltz:
```suggestion
const B_METAFEED = Buffer.from('metafeed')
```

lookup.js
```diff
@@ -49,6 +49,22 @@ function subfeed(feedId) {
...
+  if (p < 0) return
+  p = seekKey(buffer, p, B_CONTENT)
+  if (p < 0) return
+  return seekKey(buffer, p, B_PARENT)
+}
+function parent(feedId) {
+  return equal(seekParentfeed, feedId, {
+    prefix: 32,
+    prefixOffset: PARENTFEED_PREFIX_OFFSET,
+    indexType: 'value_content_parent',
```
staltz, line 64:
```suggestion
    indexType: 'value_content_metafeed',
```

lookup.js
```diff
@@ -289,6 +305,27 @@ exports.init = function (sbot, config) {
...
+  function isRootFeedId(feedId, cb) {
+    try {
+      assertFeedId(feedId)
+      if (!validate.detectFeedFormat(feedId)) throw Error('Invalid feedId')
+    } catch (err) {
+      return cb(err)
+    }
+
+    sbot.db.query(
+      where(parent(feedId)),
```
staltz, line 317: I'm not sure this is the appropriate indicator of being a root metafeed. In `query.js` (by the way, why is this PR for lookup.js and not query.js?) we use `getAnnounces()` to discover root MFs, and I think that's a more sure way. Because it's not hard for a developer to randomly use `'v1'` as a feedpurpose somewhere for their own leaf, but then suddenly this function would think that e.g. a shard is a root.

staltz (changes requested):

lookup.js
Powersource: @mixmix

mixmix: I put it in `lookup.js` because it felt like the analogue to the lookup in that was happening in the other direction. Happy to move it if there's somewhere more sensible.

Will look into your suggestion about `getAnnounces` and come back

mixmix: have moved to use @staltz 's suggestion, and put it in `query.js`

This will need to change if we ever decide to drop the `main` feed, but can figure that out later

mixmix:
@Powersource this is ready to go. I'm merging it once tests pass

mixmix:
GREAT. CI is refusing to run because staltz requested changes.... but I resolved them... but there is a bug. NICE

mixmix:
sigh... CI was set to only run tests on PRs to main.
Just gonna merge

PR #121, opened 2023-02-17
add ability to skipMainFeed connection

mixmix:
## Background

`ssb-tribes2` test suite was taking 3min 40s to run on my desktop. I did a little analysis and found that there was sometimes significant slow-downs calling `ssb.metafeeds.findOrCreate` , in particular the `ssb.tribes.start()` seems to take 3 seconds to complete (and it creates locks).

## Improvement

I noticed that unless we were testing persistence, setting up the `main` feed (publishing the rootMetaFeedId, and setting the `main` feed up in a shard...) is slow, and not actually useful for most tests.

I experimented with extracting that step and making it possible to turn it off.

### DIFF
--- a/api.js
+++ b/api.js
@@ -213,35 +213,40 @@ exports.init = function (sbot, config = {}) {
       }
       sbot.metafeeds.lookup.updateMyRoot(mf)
 
-      // Ensure root meta feed announcement exists on the main feed
-      const [err2, announcements] = await run(getAnnounces)()
-      if (err2 || !announcements || announcements.length === 0) {
-        if (err2) debug('announcing meta feed on main feed because %o', err2)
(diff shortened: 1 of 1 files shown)


mixmix:
Here it is in `ssb-tribes2`

staltz:
Profiling shows that it takes about 500ms to announce the root on the main, but it takes 1600ms to create `/root/v1/:shard/main`.

staltz:
I think all of these waits are coming from lodash.debounce which is sprinkled around in ssb-db2 and async-append-only-log. In some tests we have disabled those timers so that tests run quicker.

staltz:
Can we close this PR? @mixmix

PR #122, opened 2023-04-05
Add test for branchStream root opt

### LINKED ISSUE #114: branchstream with root doesn't seem to work
I'm trying to fix this [unpin ssb-meta-feeds](https://github.com/ssbc/ssb-tribes2/issues/48) again in tribes2 and it seems that there is something wrong in `branchstream` when you pass it a root. This is also not tested as far as I can see. What I'm seeing is that is just returns your whole forest. Instead of just for that root. And this [line](https://github.com/ssbc/ssb-meta-feeds/blob/08ef0edeff00fa9bf3a1f9b9b6bd77589647e859/lookup.js#L90) looks wrong. It basically just tests that you have a seed yourself, not that it correspond to anything within the message. 

Anyway this is a vague issue, just wanted to put it out there. I might not have time to look at it for a few days.

Powersource:
fixes https://github.com/ssbc/ssb-meta-feeds/issues/114

* [ ] fix bug where you can't use root on someone else's root id
* [ ] add check that the user is actually inputting a root id and not like an object
* [ ] shouldn't output anything if it's an invalid root id. like here it just seems to make up a root feed https://github.com/ssbc/ssb-tribes2/pull/78#discussion_r1158493143
* [ ] arj wrote that "this [line](https://github.com/ssbc/ssb-meta-feeds/blob/08ef0edeff00fa9bf3a1f9b9b6bd77589647e859/lookup.js#L90) looks wrong. It basically just tests that you have a seed yourself, not that it correspond to anything within the message. "

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "husky": "^4.3.0",
     "prettier": "^2.1.2",
     "pretty-quick": "^3.1.0",
+    "pull-many": "^1.0.9",
     "rimraf": "^3.0.2",
     "secret-stack": "^6.4.0",
     "ssb-bendy-butt": "^1.0.1",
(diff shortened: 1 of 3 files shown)


Powersource:
anyone want to adopt this pr? it seems i won't need it for my tribes2 pr
@arj03 @mixmix @staltz

